PR #235, opened 2012-02-23
Historic slug conflicts

erikogan:
We added a fix for the failing test submitted by @earnold. It patches the history plugin by adding custom SlugGenerator overrides. When querying for slug conflicts, it checks the Slug model instead of the sluggable model.

And here's a picture of an adorable baby polar bear to make up for the previous bad pull request:
Knut the cute polar bear

### DIFF
--- a/lib/friendly_id/history.rb
+++ b/lib/friendly_id/history.rb
@@ -119,13 +119,14 @@ method.
       private
 
       def conflicts
-        pkey  = sluggable.class.primary_key
-        value = sluggable.send pkey
+        sluggable_class = friendly_id_config.model_class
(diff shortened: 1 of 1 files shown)


Buildzzz:
Hi, I wanted to reopen this issue.  I'm using 5.2.1 and running into what I believe to be the same issue.  I turned an old topic model that use :slugged to use :history . 

Now whenever I try to create a similar topic with the subject, I get the following error:

ERROR:  duplicate key value violates unique constraint "index_forem_topics_on_slug"
DETAIL:  Key (slug)=(well-what) already exists.

Please let me know what other information you need.  

Thx!

PR #572, opened 2014-06-13
[Fix] Solved a problem caused by 'where.not' when using the globalize gem

marzapower:
I recently found out that using `friendly_id` + `globalize` will break when the app is running with postgres as database.

In particular, calling `.where.not( ... )` on a scope will append `AND (NOT ('--- id: '))` to queries, thus breaking the query on the database.

This fix converts the `.where.not()` in a generic `where()` clause through `arel` syntax.

### DIFF
--- a/lib/friendly_id/scoped.rb
+++ b/lib/friendly_id/scoped.rb
@@ -122,13 +122,12 @@ an example of one way to set this up:
     end
 
     def scope_for_slug_generator
-      arel_table = self.class.arel_table
       relation = self.class.unscoped.friendly
       friendly_id_config.scope_columns.each do |column|
         relation = relation.where(column => send(column))
       end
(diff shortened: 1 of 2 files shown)


marzapower:
Hi again. Still not released? Is there a public release plan for this kind of fixes?

norman:
No, there's no public release plan. When I planned to release this weekend but have not had the time. When 5.1.0 is pushed it will be basically identical to 5.1.0.beta.1, so you may wish to update your dependency to that version now if you need it. The only holdup right now is the fact that I want to backport the Rails 4.2 support into 5.0-stable and release that simultaneously, and I need about 2 hours to do that; 2 hours I can't seem to find right now. Thanks for your patience in the mean time.

PR #637, opened 2015-01-18
Fix slug for composite primary keys

lordmortis:
I'd like to add tests and anything else you need to get this merged? (This increases compatibility with [Composite Primary Keys](https://github.com/composite-primary-keys/composite_primary_keys))

### DIFF
--- a/lib/friendly_id/scoped.rb
+++ b/lib/friendly_id/scoped.rb
@@ -140,7 +140,7 @@ an example of one way to set this up:
       else
         whereclause = self.class.base_class.arel_table[primary_key_name].not_eq(send(primary_key_name))
       end
-      scope.where(whereclause)
+      relation.where(whereclause)
     end
     private :scope_for_slug_generator
 
--- a/lib/friendly_id/scoped.rb
+++ b/lib/friendly_id/scoped.rb
@@ -127,7 +127,20 @@ an example of one way to set this up:
         relation = relation.where(column => send(column))
       end
       primary_key_name = self.class.primary_key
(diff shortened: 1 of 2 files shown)


parndt:
I wonder if composite key support could be built as a plugin instead, to avoid all this conditional logic. For example:

```ruby
friendly_id :title, use: [:slugged, :composite_key]
```

lordmortis:
@parndt would that make testing easier? I kind-of-like the "just works" approach, although you do have to be careful about your include ordering to get friendly Id and CompositePrimaryKeys working together (because they both override the param method)

parndt:
No, I just think the patch is quite verbose and makes the code more difficult to understand. The fact that it's repeated for scoped and slugged makes me think it should be its own thing.

norman:
I'd prefer not to include this - if this were intended to support a feature that's in Active Record then it would be fine, but I don't think the core of FriendlyId should include support for 3rd libraries. That should go in its own module as @parndt suggested. This is the approach we've taken with Globalize, where friendyld_id-globalize is its own gem.

However if there are modifications needed to make FriendlyId easier to extend and make such a module/gem possible, then I'd be happy to add that. Perhaps the various modules need to provide a `primary_key_scope` method that can be overridden where necessary.

lordmortis:
[Here](/lordmortis/friendly_id-composite_primary_keys) <- that's my first pass at the plugin (which is basically just overriding the scope_for_slug_generator like my patch did. 

primary_key_scope would likely make this a lot easier to do, and i'll look at that when i get some more time. For now, if anyone is looking to use friendly_id with composite_primary_key - there's now a plugin! (not sure if there's a place these get listed?)

PR #644, opened 2015-02-09
Addon for numerically sequential slugs

tekin:
This provides an addon that enables sequential slugs for conflict resolution, mimicking the behaviour of friendly_id v4.

Well formed candidates combined with UUIDs as a fallback will almost always be the better solution for resolving slug conflicts. That said, there are circumstances when this is either not possible or undesirable. For example, either good candidates cannot be formed, or the existing behaviour needs to be maintained for consistency.

I have recently had to deal with this situation whilst upgrading a [large Rails 3 application to Rails 4](https://github.com/alphagov/whitehall/pull/1958). In this case it's not possible to form appropriate candidates for many of the sluggable models, and UUIDs would make new URLs inconsistent and "ugly". Ultimately it's both cheaper and far easier for the business (in this case the [UK government](https://github.com/alphagov)) to maintain the existing behaviour than it would be to retrospectively migrate to the candidate/UUID-based approach.

I have intentionally left out any documentation pending feedback. I'm assuming that if this is to make it into friendly_id, it would have to be made clear that it's meant as a last resort for those unable to use the preferred candidate/UUID-based approach.

/cc @parndt

### DIFF
--- a/lib/friendly_id/sequentially_slugged.rb
+++ b/lib/friendly_id/sequentially_slugged.rb
@@ -1,0 +1,67 @@
+module FriendlyId
+  module SequentiallySlugged
+    def self.setup(model_class)
+      model_class.friendly_id_config.use :slugged
+    end
+
+    def should_generate_new_friendly_id?
+      send(friendly_id_config.base).present? && super
+    end
+
+    def resolve_friendly_id_conflict(candidate_slugs)
+      SequentialSlugCalculator.new(scope_for_slug_generator,
+                                  candidate_slugs.first,
+                                  friendly_id_config.slug_column,
+                                  friendly_id_config.sequence_separator).next_slug
+    end
+
+    class SequentialSlugCalculator
+      attr_accessor :scope, :slug, :slug_column, :sequence_separator
+
+      def initialize(scope, slug, slug_column, sequence_separator)
+        @scope = scope
+        @slug = slug
+        @slug_column = scope.connection.quote_column_name(slug_column)
+        @sequence_separator = sequence_separator
+      end
+
+      def next_slug
+        slug + sequence_separator + next_sequence_number.to_s
+      end
+
+    private
+
+      def next_sequence_number
+        if last_sequence_number == 0
+          2
+        else
+          last_sequence_number + 1
+        end
+      end
+
+      def last_sequence_number
+        slug_conflicts.last.split("#{slug}#{sequence_separator}").last.to_i
+      end
+
+      def slug_conflicts
+        scope.
+          where(conflict_query, slug, sequential_slug_matcher).
+          order(ordering_query).pluck(slug_column)
+      end
+
+      def conflict_query
+        "#{slug_column} = ? OR #{slug_column} LIKE ?"
+      end
+
+      def sequential_slug_matcher
+        "#{slug}#{sequence_separator}%"
+      end
+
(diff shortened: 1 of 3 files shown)


norman:
I'm on vacation without my computer so I can't give good feedback on this for at least a week, but I'm in favor of this feature and grateful that you worked on it. When we had this feature by default, I remember there bein so fairly complex code to pull the sequence off the end of the slug in order to increment it, and to extract the base for SQL queries. You should make sure to look through the git and issue history for this feature in version 4.x so we don't end up repeating the cycle of fixing bugs that were solved years before. 

Your suggestions for documentation sound perfect. Thanks again for taking this on!

parndt:
@tekin what a legend

lib/friendly_id/sequentially_slugged.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+
+      def sequential_slug_matcher
+        "#{slug}#{sequence_separator}%"
+      end
+
+      # Return the unnumbered (shortest) slug first, followed by the numbered ones
+      # in ascending order.
+      def ordering_query
+        "LENGTH(#{slug_column}) ASC, #{slug_column} ASC"
```
parndt: does this work in sqlite3, postgresl, mysql etc?

tekin: I've only tested it with MySQL myself, but according to the Travis build it does: https://travis-ci.org/norman/friendly_id/builds/50012697

tekin:
Cheers guys. The code is mostly based on the previous slug generator code from v4, but there are a few bits I didn't port across because mainly because I couldn't get my head around the failing case they were trying to solve. I should have some time later in the week to go back through it and see if there are any edge cases that I've failed to cover.

gaganawhad:
Thanks for your work @tekin

tekin:
I've increased the test coverage and made a change to ensure the slug column is correctly escaped in the queries used to generate sequential slugs.

There is a particular bit of code from the previous implementation that I've not ported: https://github.com/norman/friendly_id/pull/244/files - I *think* I understand the issue this is trying to solve, but I've really struggled to replicate a failing test. This makes me reluctant to simply copy across the code that is supposedly doing the required escaping.

tekin:
Hi @norman anything I can do to help move this forward. I'm not entirely sure what the failing build is about. It appears unrelated.

norman:
@tekin sorry for the delay, I'll try to make time for this soon and get it (and the other PRs that are accumulating) merged in.

mtarnovan:
@norman Any idea when this will be merged? The build doesn't seem to fail from this PR. Thanks.

norman:
Sorry again for my long absence. I'm working on this today and will get it merged in along with the other open PRs.

norman:
@tekin I picked through the history of this functionality in version 4.0 and added in support to escape the '%' and '_' characters in LIKE queries, the 'LEN' operator for MSSQL and specifying `ESCAPE '\\'` for SQLite3 - all issues that came up one time or another in the past. I've now merged your PR into master. Thanks again for your excellent work, and your patience as I've struggled to get the time and focus to give this the attention it deserved.

tekin:
Thanks @norman, really appreciate your time and effort in creating and maintaining the gem :cake:

norman:

PR #651, opened 2015-03-04
Make Candidates#each iterable without block.

ignat-zakrevsky:
Hi

The mean aim of this commit is to provide access to generated slug candidates without direct AR manipulations
```ruby 
FriendlyId::Candidates.new(model, model.slug_candidates).each 
# => ["ergonomic-zero-es-africa", "ergonomic-zero-es-africa-100"]
# instead of 
# LocalJumpError: no block given (yield)
``` 
So, it seems reasonable for me to use this syntax instead of calling #save. It can be written to return Enumerator instead of array, but it will be not so readable and will complicate method.

### DIFF
--- a/lib/friendly_id/candidates.rb
+++ b/lib/friendly_id/candidates.rb
@@ -23,6 +23,9 @@ module FriendlyId
       unless pre_candidates.all? {|x| reserved?(x)}
         pre_candidates.reject! {|x| reserved?(x)}
       end
+
+      return pre_candidates unless block_given?
+
       pre_candidates.each {|x| yield x}
     end
 
--- a/test/candidates_test.rb
+++ b/test/candidates_test.rb
@@ -114,4 +114,30 @@ class CandidatesTest < TestCaseClass
     end
   end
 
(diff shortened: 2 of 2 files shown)


parndt:
Oh, I like this!  We'll need tests (please :heart:) to ensure that the functionality works like you say it does now and in the future, though. :)

ignat-zakrevsky:
Ok, no problem

ignat-zakrevsky:
@parndt, updated

parndt:
@norman over to you. 

FYI, CI is (and was before this patch) failing due to ActiveRecord 4.2 + postgresql + jruby.. I had a very similar / the same thing happen in [awesome_nested_set (commit)](https://github.com/collectiveidea/awesome_nested_set/commit/783b41533c4927c152175a5bb8504e80bffbb7e9), though I'm not exactly sure why it fails here.

Here's the [awesome_nested_set job](https://travis-ci.org/collectiveidea/awesome_nested_set/jobs/52845976)

norman:
Merged, thanks!

norman:

PR #667, opened 2015-05-23
Don't calculate all changes just to check if the param field has changed

mhodgson:
Starting with Rails 4.2, calculating the `changes` hash in activerecord is much more expensive since it now tracks in-place changes in json/jsonb/hstore fields. We noticed a ~10x hit to performance due to these changes being calculated on every `to_param` call on some of our models with large json fields. 

This change simply checks for changes in a less expensive way before calculating the `changes` hash if the param field has indeed changed. 

All tests still pass and I didn't see a need to add additional testing.


norman:
Merged, thanks!

mhodgson:
Of course, thanks for getting it merged quickly!

norman:

PR #671, opened 2015-05-29
add GoRails video to readme documentation

sublimecoder:
Add go rails video to documentation and fix styling issue with the two links in articles being on the same line.


norman:
Thanks for this! Maybe you could put the gorails video before the Railscast
one, since it's much newer?

sublimecoder:
Sure thing, Just moved it and change the copy to reflect that change.

norman:
Thanks!

PR #674, opened 2015-06-07
Possibly optimize the unfriendly_id? check

dgynn:
The FinderMethods' find() does an `unfriendly_id?` check on the `id` before performing the query.
The logic for that check is a little complex since it also handles the unknown (nil) case.

It seems that the primary cases that need to be handled are String and Numeric. This commit
fast-paths those checks so that Strings always are alway not unfriendly and are checked by the finder and Numerics always immediately pass on to `super`. All `unfriendly_id?` checks are done without allocating any new objects.

This commit changes the monkey-patching of Object, which of course isn't great, but that door has already been opened. Another option would be to stop using `unfriendly_id?` as the finder check condition and implement a new method optimized for Strings and Numerics.

Here are the benchmarks of the new approach. Two things to note...
* This is for 5 million iterations. Relative to the finder `find()`, this is only a 3-5% improvement against a local DB and slight memory improvement.
* In the existing code, the order of a class in the `UNFRIENDLY_CLASSES` array impacts speed. So Numeric could be moved first even if no other change was made.

Before
```
------------------------------------------------------------------------
Using ruby 2.2.2 AR 4.2.1 with sqlite3 (in-memory)
Rehearsal -----------------------------------------------------------------
integer friendly_id?            6.530000   0.010000   6.540000 (  6.533914)
integer unfriendly_id?          6.780000   0.000000   6.780000 (  6.788747)
AR::Base friendly_id?           2.290000   0.000000   2.290000 (  2.284586)
AR::Base unfriendly_id?         2.560000   0.000000   2.560000 (  2.568194)
hash friendly_id?               5.240000   0.010000   5.250000 (  5.243626)
hash unfriendly_id?             5.460000   0.000000   5.460000 (  5.460569)
nil friendly_id?                5.690000   0.000000   5.690000 (  5.697026)
nil unfriendly_id?              5.980000   0.010000   5.990000 (  5.985974)
numeric string friendly_id?     8.640000   0.000000   8.640000 (  8.645938)
numeric string unfriendly_id?   9.620000   0.020000   9.640000 (  9.661533)
test_string friendly_id?        8.290000   0.000000   8.290000 (  8.292507)
test_string unfriendly_id?      8.450000   0.010000   8.460000 (  8.456376)
------------------------------------------------------- total: 75.590000sec

                                    user     system      total        real
integer friendly_id?            6.400000   0.000000   6.400000 (  6.401500)
integer unfriendly_id?          6.740000   0.000000   6.740000 (  6.736401)
AR::Base friendly_id?           2.220000   0.010000   2.230000 (  2.218202)
AR::Base unfriendly_id?         2.510000   0.000000   2.510000 (  2.518849)
hash friendly_id?               5.060000   0.000000   5.060000 (  5.067301)
hash unfriendly_id?             5.420000   0.000000   5.420000 (  5.420728)
nil friendly_id?                5.550000   0.000000   5.550000 (  5.551557)
nil unfriendly_id?              5.940000   0.010000   5.950000 (  5.947000)
numeric string friendly_id?     8.400000   0.000000   8.400000 (  8.405858)
numeric string unfriendly_id?   8.800000   0.000000   8.800000 (  8.800442)
test_string friendly_id?        8.080000   0.010000   8.090000 (  8.082986)
test_string unfriendly_id?      9.270000   0.020000   9.290000 (  9.308659)
```
After
```
Using ruby 2.2.2 AR 4.2.1 with sqlite3 (in-memory)
Rehearsal -----------------------------------------------------------------
integer friendly_id?            0.300000   0.000000   0.300000 (  0.302883)
integer unfriendly_id?          0.320000   0.000000   0.320000 (  0.313744)
AR::Base friendly_id?           0.290000   0.000000   0.290000 (  0.299272)
AR::Base unfriendly_id?         0.310000   0.000000   0.310000 (  0.302496)
hash friendly_id?               0.320000   0.000000   0.320000 (  0.319220)
hash unfriendly_id?             0.310000   0.000000   0.310000 (  0.315932)
nil friendly_id?                0.300000   0.000000   0.300000 (  0.300949)
nil unfriendly_id?              0.310000   0.000000   0.310000 (  0.313623)
numeric string friendly_id?     1.660000   0.000000   1.660000 (  1.667411)
numeric string unfriendly_id?   0.320000   0.000000   0.320000 (  0.319070)
test_string friendly_id?        1.330000   0.000000   1.330000 (  1.331005)
test_string unfriendly_id?      0.310000   0.000000   0.310000 (  0.309833)
-------------------------------------------------------- total: 6.080000sec

                                    user     system      total        real
integer friendly_id?            0.340000   0.000000   0.340000 (  0.332969)
integer unfriendly_id?          0.310000   0.000000   0.310000 (  0.311426)
AR::Base friendly_id?           0.310000   0.000000   0.310000 (  0.311694)
AR::Base unfriendly_id?         0.320000   0.000000   0.320000 (  0.319094)
hash friendly_id?               0.330000   0.000000   0.330000 (  0.321552)
hash unfriendly_id?             0.300000   0.000000   0.300000 (  0.304965)
nil friendly_id?                0.320000   0.000000   0.320000 (  0.319326)
nil unfriendly_id?              0.330000   0.000000   0.330000 (  0.325399)
numeric string friendly_id?     1.650000   0.000000   1.650000 (  1.650626)
numeric string unfriendly_id?   0.310000   0.000000   0.310000 (  0.311350)
test_string friendly_id?        1.330000   0.010000   1.340000 (  1.336598)
test_string unfriendly_id?      0.310000   0.000000   0.310000 (  0.311427)
```

### DIFF
--- a/test/object_utils_test.rb
+++ b/test/object_utils_test.rb
@@ -1,7 +1,5 @@
 require "helper"
 
-using FriendlyId::ObjectUtils
-
 class ObjectUtilsTest < TestCaseClass
 
   include FriendlyId::Test
--- a/test/object_utils_test.rb
+++ b/test/object_utils_test.rb
@@ -8,21 +8,38 @@ class ObjectUtilsTest < TestCaseClass
 
   test "strings with letters are friendly_ids" do
     assert "a".friendly_id?
+    assert "a".possibly_friendly_id?
+    refute "a".unfriendly_id?
   end
 
   test "integers should be unfriendly ids" do
+    refute 1.friendly_id?
+    refute 1.possibly_friendly_id?
     assert 1.unfriendly_id?
   end
 
-  test "numeric strings are neither friendly nor unfriendly" do
-    assert_equal nil, "1".friendly_id?
-    assert_equal nil, "1".unfriendly_id?
+  test "numeric strings are neither friendly nor unfriendly but are possibly friendly" do
+    assert_nil "1".friendly_id?
+    assert "1".possibly_friendly_id?
+    assert_nil "1".unfriendly_id?
   end
 
   test "ActiveRecord::Base instances should be unfriendly_ids" do
     model_class = Class.new(ActiveRecord::Base) do
       self.table_name = "authors"
     end
+    refute model_class.new.friendly_id?
+    refute model_class.new.possibly_friendly_id?
     assert model_class.new.unfriendly_id?
   end
+
+  test "any object that responds to to_i and to_s" do
+    TestClass=Struct.new(:to_i,:to_s)
+    numericish_string_object = TestClass.new(123, "123abc")
+
+    assert numericish_string_object.friendly_id?
+    assert numericish_string_object.possibly_friendly_id?
+    refute numericish_string_object.unfriendly_id?
+  end
+
 end
--- a/test/object_utils_test.rb
+++ b/test/object_utils_test.rb
@@ -6,19 +6,16 @@ class ObjectUtilsTest < TestCaseClass
 
   test "strings with letters are friendly_ids" do
     assert "a".friendly_id?
-    assert "a".possibly_friendly_id?
     refute "a".unfriendly_id?
   end
 
   test "integers should be unfriendly ids" do
     refute 1.friendly_id?
-    refute 1.possibly_friendly_id?
     assert 1.unfriendly_id?
   end
 
-  test "numeric strings are neither friendly nor unfriendly but are possibly friendly" do
+  test "numeric strings are neither friendly nor unfriendly" do
     assert_nil "1".friendly_id?
-    assert "1".possibly_friendly_id?
     assert_nil "1".unfriendly_id?
   end
 
@@ -27,7 +24,6 @@ class ObjectUtilsTest < TestCaseClass
       self.table_name = "authors"
     end
     refute model_class.new.friendly_id?
-    refute model_class.new.possibly_friendly_id?
     assert model_class.new.unfriendly_id?
   end
 
@@ -36,7 +32,6 @@ class ObjectUtilsTest < TestCaseClass
     numericish_string_object = TestClass.new(123, "123abc")
 
     assert numericish_string_object.friendly_id?
-    assert numericish_string_object.possibly_friendly_id?
     refute numericish_string_object.unfriendly_id?
   end
 
--- a/test/object_utils_test.rb
+++ b/test/object_utils_test.rb
@@ -13,9 +13,9 @@ class ObjectUtilsTest < TestCaseClass
     assert 1.unfriendly_id?
   end
 
-  test "numeric strings are neither friendly nor unfriendly" do
+  test "numeric strings are not friendly but are unfriendly" do
     assert_equal nil, "1".friendly_id?
-    assert_equal nil, "1".unfriendly_id?
+    assert_equal false, "1".unfriendly_id?
   end
 
   test "ActiveRecord::Base instances should be unfriendly_ids" do
--- a/test/object_utils_test.rb
+++ b/test/object_utils_test.rb
@@ -1,10 +1,10 @@
 require "helper"
 
+using FriendlyId::ObjectUtils
 
 class ObjectUtilsTest < TestCaseClass
 
   include FriendlyId::Test
-  using FriendlyId::ObjectUtils
 
   test "strings with letters are friendly_ids" do
     assert "a".friendly_id?
--- a/test/object_utils_test.rb
+++ b/test/object_utils_test.rb
@@ -4,6 +4,7 @@ require "helper"
 class ObjectUtilsTest < TestCaseClass
 
   include FriendlyId::Test
+  using FriendlyId::ObjectUtils
 
   test "strings with letters are friendly_ids" do
     assert "a".friendly_id?
@@ -13,9 +14,9 @@ class ObjectUtilsTest < TestCaseClass
     assert 1.unfriendly_id?
   end
 
-  test "numeric strings are not friendly but are unfriendly" do
+  test "numeric strings are neither friendly nor unfriendly" do
     assert_equal nil, "1".friendly_id?
-    assert_equal false, "1".unfriendly_id?
+    assert_equal nil, "1".unfriendly_id?
   end
 
   test "ActiveRecord::Base instances should be unfriendly_ids" do
--- a/friendly_id.gemspec
+++ b/friendly_id.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.require_paths     = ["lib"]
   s.license           = 'MIT'
 
-  s.required_ruby_version = '>= 2.0.0'
+  s.required_ruby_version = '>= 1.9.3'
 
   s.add_dependency 'activerecord', '>= 4.0.0'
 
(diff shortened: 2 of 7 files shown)


test/object_utils_test.rb
```diff
@@ -13,9 +13,9 @@ class ObjectUtilsTest < TestCaseClass
     assert 1.unfriendly_id?
   end
 
-  test "numeric strings are neither friendly nor unfriendly" do
+  test "numeric strings are not friendly but are unfriendly" do
     assert_equal nil, "1".friendly_id?
-    assert_equal nil, "1".unfriendly_id?
+    assert_equal false, "1".unfriendly_id?
```
norman: The three-way logic was actually intentional here, and we shouldn't change it. This is because a string with nothing but integers may in fact be a friendly id. For example, assume a "Car" model that uses the model name as a slug - a Peugeot 206 may have "206" as its slug.

norman:
@dgynn Thanks a lot for your contributions! 

This looks good - I don't think this will have a huge real-world impact but it's an improvement with no more complexity, so I see no reason not to pull. Please consider my comment above about the change in the three way logic. If it turns out that changing your `StringUtils` to work exactly like the current version of FriendlyId  eliminates these performance gains, then it might not be worth making these changes.

Regarding your comment about monkey-patching I agree that it's something to be done with great caution. The reason I allow this here is because the method we're adding to `Object` has a name that's so specific to the library that it's unlikely to conflict with anything. In seven years of maintaining this library nobody has every brought it up as a real-world issue affecting their application. 

But perhaps now that Ruby 1.9 is dropping off the radar, we could consider implementing it as a refinement. I haven't even looked closely enough at refinements to know if it's a good idea, but if you'd like to look into it I'd be more than happy to consider it.

test/object_utils_test.rb
dgynn: Yeah. I didn't like changing this. The trouble is that the finder is only looking for `unfriendly_id? == true`. That is never true for any String so doing the `.to_i.to_s` check is just being ignored.
One option would be to change finder to call a different method like `definitely_not_friendly?` but I didn't want to add more things to Object. Possibly with refinements this wouldn't be an issue.

dgynn:
Great. I've tried implementing this with refinements and it seems to work in Ruby 2.1 and 2.2 but not 2.0. I'm still working on that but I wanted to push this up anyway.
If this works out then adding a `definitely_not_friendly?` check for the finder to use might not be so bad.

norman:
> If this works out then adding a definitely_not_friendly? check for the finder to use might not be so bad.

Why not check the value of `foo.friendly_id? == false`?

dgynn:
Argh. Refinements don't work in Rubinius. I'm going to switch this back to the monkey-patch way and keep the possibly_friendly_id? check.

dgynn:
Thanks for bearing with me on this... I've updated this branch to remove String logic and just patch up the Unfriendly objects. The gain from that change helps the String check too and the complexity introduced to eliminate 1 string allocation isn't worth it.

If you think this looks OK I'll squash it down and clean it up.

With this change the impact of having finders attached is reduced when doing a find() with an integer. Here are some comparisons of .find() calls with and without FinderMethods and using integer ids, string ids, and string slugs. There is a lot of variability in timings across runs. But the difference between calling find(integer) with and without FinderMethods is always reduced.

Before
```
Comparing .find() with and without Finders and using integer ids, string ids, and string slugs
Rehearsal ---------------------------------------------------------------------
Unfriendly: find(id integer)        2.760000   0.000000   2.760000 (  2.759420)
FriendlyFinder: find(id integer)    2.920000   0.010000   2.930000 (  2.920976)
Unfriendly: find(id string)         3.350000   0.000000   3.350000 (  3.353196)
FriendlyFinder: find(id string)     6.450000   0.010000   6.460000 (  6.461136)
FriendlyFinder: find(slug string)   3.420000   0.000000   3.420000 (  3.417653)
----------------------------------------------------------- total: 18.920000sec

                                        user     system      total        real
Unfriendly: find(id integer)        2.750000   0.000000   2.750000 (  2.745454)
FriendlyFinder: find(id integer)    2.890000   0.010000   2.900000 (  2.893111)
Unfriendly: find(id string)         3.270000   0.000000   3.270000 (  3.279130)
FriendlyFinder: find(id string)     6.530000   0.000000   6.530000 (  6.538249)
FriendlyFinder: find(slug string)   3.420000   0.010000   3.430000 (  3.418126)
```
After
```
Rehearsal ---------------------------------------------------------------------
Unfriendly: find(id integer)        2.620000   0.000000   2.620000 (  2.627494)
FriendlyFinder: find(id integer)    2.690000   0.000000   2.690000 (  2.689371)
Unfriendly: find(id string)         3.270000   0.000000   3.270000 (  3.278969)
FriendlyFinder: find(id string)     6.300000   0.010000   6.310000 (  6.294359)
FriendlyFinder: find(slug string)   3.180000   0.000000   3.180000 (  3.187919)
----------------------------------------------------------- total: 18.070000sec

                                        user     system      total        real
Unfriendly: find(id integer)        2.660000   0.000000   2.660000 (  2.669838)
FriendlyFinder: find(id integer)    2.680000   0.000000   2.680000 (  2.683657)
Unfriendly: find(id string)         3.220000   0.000000   3.220000 (  3.224202)
FriendlyFinder: find(id string)     6.270000   0.010000   6.280000 (  6.270034)
FriendlyFinder: find(slug string)   3.210000   0.000000   3.210000 (  3.213826)
```

kimrgrey:
Hi, @dgynn! Are you still interested in this PR? I'll be really glad to merge it after few questions!

dgynn:
Hi, @kimrgrey. I'd still be interested in getting this merged. I can do another PR of this squashed down if needed. Just let me know. Thanks!

norman:
I'm :-1: on this. I don't think the very small gain in performance justifies and added complexity, or having FriendlyId stick its fingers into so many more classes.

dgynn:
Yep. No problem. I was also not thrilled with modifying other classes. Thanks for taking a look.

kimrgrey:
Oh, I liked some parts of this PR :cry: 

@norman, you are right, my main question was: "Are we really need to change all this files in one PR?". But, as for me, idea to replace conditional logic that detects inheritance with simple module mixing in "unfrienldy" classes was nice and very-very clean. 

May be we can take 2 changes from this PR:

1. Move definition of `FriendlyId::UNFRIENDLY_CLASSES`. We are using this constant in `FriendlyId::ObjectUtils` and only here, but by some reason it's defined somewhere else. Idea to move it to the same file looks reasonable.
2. Use approach with mixing of `UnfriendlyUtils` to "unfriendly classes". In this case we will do complex checks only for something that possibly can be "friendly", and definitely "unfriendly" classes will just return false.

Simple. Useful. No need in changing of tests. What do you think?

norman:
In my view it's less simple than what's currently there and only improves performance a little. My preference would be to leave it as is. But since you're the person who is most motivated to do actual work on the library right now, I'm happy to leave it to your best judgement @kimrgrey. :smile:

kimrgrey:
@norman, cool!  I really appreciate your confidence :wink: Thanks!

@dgynn, can I use your ideas to implement new PR, that will change this two things? Or you are interested in doing it by yourself?

dgynn:
Let me push up another PR that gets most of the performance improvements without mixing in to other classes. @kimrgrey, I'm fine with having you make any changes to either approach.

Issue #602, opened 2014-09-03
FriendlyId 5: how to get old-style incrementally increasing slugs?


riffraff:
I am trying to upgrade an app to rails 4, and it depends on friendly_id, since v5 is the only one supporting rails 4 I have to uograde it as well. 

I  do not want the new "append UUID" behaviour on name conflicts, is there a way to keep the old behavior (incremental number), or should I just fork friendly_id v4 ?

Issue #576, opened 2014-06-27
Create Shorter Conflict Slugs and Recreate Slugs on Update


harshamv:
If i have a conflicting slug it generates a very long slug as follows

Explore     explore 
Explore     explore-7a8411ac-5af5-41a3-ab08-d32387679f2b

Is there a way i can just create 'explore-1', 'explore-2' etc.

Other option is i want to have a shorter slug generator like the following

candidates.first + friendly_id_config.sequence_separator + SecureRandom.hex(3)

How can i create a Initialization to allow this across all models?
Also how can i recreate the slug on updating a record in the model?

Issue #480, opened 2013-11-11
Supporting sequence numbers for conflicting slugs in friendly_id 5 


gaganawhad:
Thanks for all the work on friendly_id 5

While slug candidates and UUID both make for good solutions to resolve conflicts in a slug, I would like to use sequence numbers in slugs ('--2', '--3' etc.) in cases of conflict (which is the default solution used in friendly_id 4). 

I do not have any other reliable slug candidate than the one I am using and UUIDs are a bit too 'unfriendly' for my application. 

I would really appreciate any pointers on how I can go about doing that. If I can contribute to documentation / code for this, I will. 

p.s. While this issue may not qualify for a bug, I thought of creating an issue instead of an SO question because I think this is more tightly integrated with the software.

dortonway:
Why can't we just use id instead of sequence or UUID to have prettier urls?

norman:
> Why can't we just use id instead of sequence or UUID to have prettier urls?

You can certainly do that if you want to, but not everybody does. If you want to use the id in the URL you don't need FriendlyId - a method like this in your model is sufficient:

```ruby
def to_param
  "#{id}-#{some_method.parameterize}"
end
```

Issue #430, opened 2013-08-08
Duplicate entry generate slugs with unfriendly hash instead of incremental number


IRMooBear:
Version: 5.x  Only (4.x was working fine)

Duplicate entries in the field used to generate slug is generating an extra long and unfriendly slug with a GUID like ID appended to the end of the slug instead of an incremental number in V5.

Example, multiple database field with duplicate values
Primary Review Group
Primary Review Group
Primary Review Group
Primary Review Group

Generating the following slugs
primary-review-group
primary-review-group-8a54a6cc-72ae-4646-82fc-cc92baf2eec6
primary-review-group-4c52971c-8225-49df-b2d3-b7eac5da522c
primary-review-group-5c382dee-ba14-4f75-bced-a1807f276de2

The first record generate the correct slug, record 2+ generate something unfriendly.  4.x would have generated something like
primary-review-group-2
primary-review-group-3
primary-review-group-4

But now instead of a friendly incremental number, it's saving what appear to be some sort of unique hash.

PR #677, opened 2015-06-20
Update README.md

onerinas:
Read me update with code for previously used numeric sequence

Suggestion reference: https://github.com/norman/friendly_id/issues/602#issuecomment-54272912

Code reference: https://github.com/norman/friendly_id/issues/576#issuecomment-47485811

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,28 @@ The most important changes are:
   sequence to differentiate conflicting slug, but rather a UUID (e.g. something
   like `2bc08962-b3dd-4f29-b2e6-244710c86106`). This makes the
   codebase simpler and more reliable when running concurrently, at the expense
-  of uglier ids being generated when there are conflicts.
+  of uglier ids being generated when there are conflicts. 
+* If you need previously used numeric sequence for some reason:
+
+    ```ruby
+    class Restaurant < ActiveRecord::Base
+      extend FriendlyId
+      friendly_id :slug_candidates, use: :slugged
+      def slug_candidates
+        [ :name ]
+      end
+      
+      def slug_candidates_with_sequence
+        2.upto(10).inject([name]) { |m, o| m << "%s %d" % [name, o] }
+      end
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,7 +87,28 @@ The most important changes are:
   sequence to differentiate conflicting slug, but rather a UUID (e.g. something
   like `2bc08962-b3dd-4f29-b2e6-244710c86106`). This makes the
   codebase simpler and more reliable when running concurrently, at the expense
-  of uglier ids being generated when there are conflicts.
+  of uglier ids being generated when there are conflicts. 
+* If you need previously used numeric sequence for some reason:
```
parndt: I think we should remove phrases like `for some reason` as this potentially alienates those with this requirement by making their use case seem trivial or nonsensical.  Perhaps I'm reading too far into it, but if I do then perhaps others will too. :blue_heart:

onerinas:
should we consider reframing this sentence 
or
 add a sentence like "the old behaviour is unsupported by the new code base and discouraged, but if you really think you need it check issues #430 #480, #576 and #602 for discussion and workarounds". as mentioned in https://github.com/norman/friendly_id/issues/602#issuecomment-54272912 
or 
do nothing :smile:

README.md
onerinas: should we consider re-framing this sentence to "If you want to override UUID with numeric sequence: "
or
add a sentence like "the old behaviour is unsupported by the new code base and discouraged, but if you really think you need it check issues #430 #480, #576 and #602 for discussion and workarounds". as mentioned in #602 (comment) 
or 
do nothing :smile:

kimrgrey: It looks a little bit outdated now. May be it's better to add some documentation about `SequentiallySlugged ` instead of it?

kimrgrey:
Well, I'll close this PR because we should add some documentation  about module `:sequentially_slugged` instead of it. Thank you for your contribution, anyway! We really appreciate it! :heart:

Issue #680, opened 2015-06-24
Error after I run "rails g friendly_id" on MySQL utf8mb4 / utf8mb4_unicode_ci


GabrielCabrera:
```
My Enviroment
ruby 2.2.2p95
Rails 4.2.2
mysql: stable 5.6.25 - InnoDB
friendly_id (5.1.0)
```

On a new rails app, after I run `rails g friendly_id` when I try to execute the migration, mysql sends an error of a `max key length`

    Mysql2::Error: Specified key was too long; max key length is 767 bytes: CREATE  INDEX `index_friendly_id_slugs_on_slug_and_sluggable_type`  ON `friendly_id_slugs` (`slug`, `sluggable_type`)

This only happens when I set mysql encoding to **utf8mb4** and collation **utf8mb4_unicode_ci**, because if I leave the default configuration on utf8, it runs without problems.

Any idea how to fix this please?

abookyun:
Same issue here, and this [thread on SO](http://stackoverflow.com/questions/10931307/ruby-on-rails-rake-schema-max-key-length-is-767-bytes) helped 😇

changes migration `add_index` statement like:

```ruby
    add_index :friendly_id_slugs, [:slug, :sluggable_type], length: { name: 100, slug: 20, sluggable_type: 20 }
    add_index :friendly_id_slugs, [:slug, :sluggable_type, :scope], length: { name: 100, slug: 20, sluggable_type: 20, scope: 20 }, unique: true
```

But I have no idea what length would properly fit, anyone knows?

abookyun:
More info here

Original migrated schema.rb

```ruby
add_index "friendly_id_slugs", ["slug", "sluggable_type", "scope"], name: "index_friendly_id_slugs_on_slug_and_sluggable_type_and_scope", unique: true, using: :btree
add_index "friendly_id_slugs", ["slug", "sluggable_type"], name: "index_friendly_id_slugs_on_slug_and_sluggable_type", using: :btree
add_index "friendly_id_slugs", ["sluggable_id"], name: "index_friendly_id_slugs_on_sluggable_id", using: :btree
add_index "friendly_id_slugs", ["sluggable_type"], name: "index_friendly_id_slugs_on_sluggable_type", using: :btree
```

Specified length migrated schema.rb

```ruby
[...]

PR #694, opened 2015-08-20
Specify length of multi-column index name

abookyun:
Should 🔨 #680 ⬅️ sees more discussion

See also: http://stackoverflow.com/questions/10931307/ruby-on-rails-rake-schema-max-key-length-is-767-bytes

### DIFF
--- a/lib/friendly_id/migration.rb
+++ b/lib/friendly_id/migration.rb
@@ -8,8 +8,8 @@ class CreateFriendlyIdSlugs < ActiveRecord::Migration
       t.datetime :created_at
     end
     add_index :friendly_id_slugs, :sluggable_id
-    add_index :friendly_id_slugs, [:slug, :sluggable_type], length: { name: 100, slug: 20, sluggable_type: 20 }
-    add_index :friendly_id_slugs, [:slug, :sluggable_type, :scope], length: { name: 100, slug: 20, sluggable_type: 20, scope: 20 }, unique: true
+    add_index :friendly_id_slugs, [:slug, :sluggable_type], length: { slug: 140, sluggable_type: 50 }
+    add_index :friendly_id_slugs, [:slug, :sluggable_type, :scope], length: { slug: 70, sluggable_type: 50, scope: 70 }, unique: true
     add_index :friendly_id_slugs, :sluggable_type
   end
 end
--- a/lib/friendly_id/migration.rb
+++ b/lib/friendly_id/migration.rb
@@ -8,8 +8,8 @@ class CreateFriendlyIdSlugs < ActiveRecord::Migration
       t.datetime :created_at
     end
     add_index :friendly_id_slugs, :sluggable_id
-    add_index :friendly_id_slugs, [:slug, :sluggable_type]
-    add_index :friendly_id_slugs, [:slug, :sluggable_type, :scope], :unique => true
+    add_index :friendly_id_slugs, [:slug, :sluggable_type], length: { name: 100, slug: 20, sluggable_type: 20 }
+    add_index :friendly_id_slugs, [:slug, :sluggable_type, :scope], length: { name: 100, slug: 20, sluggable_type: 20, scope: 20 }, unique: true
     add_index :friendly_id_slugs, :sluggable_type
   end
 end


DanielNill:
:+1:   I'd advocate merging this.  I ran into the same issue running Percona and I can't image this raising issues on other dbs.

norman:
I'm a bit unfamiliar with how this feature works in MySQL. Would this limit the slug column's length to 20 bytes, or make it so that it just the indexes the first 20 bytes of the row value? Also, are there any performance implications?

abookyun:
@norman According to [MySQL Document](https://dev.mysql.com/doc/refman/5.6/en/create-index.html) shows it use the first 20 bytes of the row value for index. And seems to depends on us to decide what size could *usually differentiate*. So I couldn't tell if 20 bytes is enough, but this is a way to deal with default 767 bytes limits of MySQL error.

```
The statement shown here creates an index using the first 10 characters of the name column:

CREATE INDEX part_of_name ON customer (name(10));
If names in the column usually differ in the first 10 characters, this index should not be much slower than an index created from the entire name column. Also, using column prefixes for indexes can make the index file much smaller, which could save a lot of disk space and might also speed up INSERT operations.

Prefix support and lengths of prefixes (where supported) are storage engine dependent. For example, a prefix can be up to 767 bytes long for InnoDB tables or 3072 bytes if the innodb_large_prefix option is enabled. For MyISAM tables, the prefix limit is 1000 bytes. The NDB storage engine does not support prefixes (see Section 18.1.6.6, “Unsupported or Missing Features in MySQL Cluster”).
```

norman:
@abookyun Thanks for the info. I also don't really understand the options given here. There's no "name" column on in the friendly_id_slugs table, what is the `{name: 100}` option for?

Slugs are often differentiated by the final few characters, so I think it would also make sense to index more than 20 bytes of the slug field as well. Assuming the worse-case scenario is a row with each column filled to its maximum length with 4-byte characters, perhaps this makes more sense:

```ruby
add_index :friendly_id_slugs, [:slug, :sluggable_type], length: { slug: 140, sluggable_type: 50 }
add_index :friendly_id_slugs, [:slug, :sluggable_type, :scope], length: { slug: 70, sluggable_type: 50, scope: 70 }, unique: true
```
According to the docs,

> Prefix limits are measured in bytes, whereas the prefix length in CREATE INDEX statements is interpreted as number of characters for nonbinary data types (CHAR, VARCHAR, TEXT). Take this into account when specifying a prefix length for a column that uses a multibyte character set. 

So in the above, with all 4-byte characters the largest possible indexed prefix for a row would be (140 * 4 + 50 * 4 = 760) and (70 * 4 + 70 * 4 + 50 * 4 = 760).

Again, I'm operating with only partial understanding here so please correct me if I've missed something!

abookyun:
@norman You did well at math 😉 and the fix is just pushed. thanks!

norman:
Thanks @abookyun! If you can verify that you've tried it again on MySQL and that it works, I'll merge it right now.

abookyun:
I new a dummy app with mysql as database, and the migrations log as below. I think it's worked?

```
== 20150922141644 CreateFriendlyIdSlugs: migrating ============================
-- create_table(:friendly_id_slugs)
   -> 0.1325s
-- add_index(:friendly_id_slugs, :sluggable_id)
   -> 0.0629s
-- add_index(:friendly_id_slugs, [:slug, :sluggable_type], {:length=>{:slug=>140, :sluggable_type=>50}})
   -> 0.0698s
-- add_index(:friendly_id_slugs, [:slug, :sluggable_type, :scope], {:length=>{:slug=>70, :sluggable_type=>50, :scope=>70}, :unique=>true})
   -> 0.2090s
-- add_index(:friendly_id_slugs, :sluggable_type)
   -> 0.0545s
== 20150922141644 CreateFriendlyIdSlugs: migrated (0.5293s) ===================

== 20150922141647 CreateUsers: migrating ======================================
-- create_table(:users)
   -> 0.0428s
-- add_index(:users, :slug, {:unique=>true})
   -> 0.0293s
== 20150922141647 CreateUsers: migrated (0.0723s) =============================
```

norman:
Looks like it. You light want to try creating some records to be sure
inserts and indexing is working correctly.

abookyun:
here you are 😎

```
Loading development environment (Rails 4.2.4)
irb(main):001:0> User.create(name: "alexandar the drift")
   (0.3ms)  BEGIN
  User Exists (98.2ms)  SELECT  1 AS one FROM `users` WHERE (`users`.`id` IS NOT NULL) AND `users`.`slug` = 'alexandar-the-drift' LIMIT 1
  SQL (1.1ms)  INSERT INTO `users` (`name`, `slug`, `created_at`, `updated_at`) VALUES ('alexandar the drift', 'alexandar-the-drift', '2015-09-22 15:27:05', '2015-09-22 15:27:05')
   (8.0ms)  COMMIT
=> #<User id: 1, name: "alexandar the drift", slug: "alexandar-the-drift", created_at: "2015-09-22 15:27:05", updated_at: "2015-09-22 15:27:05">
irb(main):002:0> User.friendly.find("alexandar-the-drift")
  User Load (333.2ms)  SELECT  `users`.* FROM `users` WHERE `users`.`slug` = 'alexandar-the-drift' LIMIT 1
=> #<User id: 1, name: "alexandar the drift", slug: "alexandar-the-drift", created_at: "2015-09-22 15:27:05", updated_at: "2015-09-22 15:27:05">
```

abookyun:
btw, should I worry about the CI failure?

norman:
I don't think so - it looks like an issue with our Gemfile not having the correct gems specified. It passes with MySQL on JRuby so I'd assume it's working correctly as far as MySQL is concerned. Thanks for your contribution!

abookyun:
Great! thanks @norman

abookyun:
@norman I just encounter this CI failure in our projects lol. It's caused by incompatible version of mysql2 and rails 4.x. more info: https://github.com/brianmario/mysql2/issues/675

PR #697, opened 2015-09-02
Added function to strip out possessive apostrophes

itsterry:
Our SEO types tell us that slugs with possessive apostrophes split out are bad for search scores

So logan-s-run scores worse than logans-run if you're searching for "Logan's Run"

I've added a patch (and tests) to produce the logans-run version

### DIFF
--- a/lib/friendly_id/slugged.rb
+++ b/lib/friendly_id/slugged.rb
@@ -288,7 +288,18 @@ Github issue](https://github.com/norman/friendly_id/issues/185) for discussion.
     # @param [#to_s] value The value used as the basis of the slug.
     # @return The candidate slug text, without a sequence.
     def normalize_friendly_id(value)
-      value.to_s.parameterize
+      strip_posessive_apostrophes(value).parameterize
+    end
+
+
+    # Removes single apostrophes indicating possesion
+    # Like "boy's", "girl's", etc
+    # BEFORE parameterizing, so we get
+    # boys, girls
+    # rather than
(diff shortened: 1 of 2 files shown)


cllns:
+1

Should you also look for ‘smart’ quotes?   i.e. `’`

norman:
FriendlyId relies on Rails's `to_param` method to generate slugs. This method does not remove apostrophes. I've advocated for this change in Rails itself (see rails/rails#8110), but the idea was rejected because it would break backwards compatibility.

One of my major design goals with FriendlyId is to keep it as close to the core of Rails as possible. Wherever it's possible to reuse something in Rails, I try to do so rather than reimplementing a similar but slightly different functionality here. I do this even if I don't 100% agree with the way Rails does things. I think this is the correct approach for a plugin-type library like this one.

For that reason, I don't want to make this change, even though I do agree that removing apostrophes is a good idea. If you're looking for a slug-generation library that takes SEO concerns very seriously, check out [Babosa](https://github.com/norman/babosa). This is a string processing library extracted from FriendlyId 3.0. Babosa **does** remove apostrophes, and handles all sorts of things that Rails doesn't.  It's very easy to use with FriendlyId; there's a description of how to do so in the README. Take a look and see if it might satisfy your needs.

itsterry:
Fair enough: makes a bunch of sense. My team love friendly_id, so we'll totally keep using it, although we might keep a local version with the apostrophe patch in :)

Issue #702, opened 2015-09-22
Release 5.2.0


pabloh:
I could really use the FriendlyId 4.x-style sequential slugs feature on a project right now.
Is there any chance version 5.2 will be released soon?.

norman:
There's currently no release timeline, but you could help us get it out sooner by trying the feature and letting us know how it works for you.

pabloh:
@norman, I've been using firendly_id master in production for a while as you suggested without noticing any problems (besides what I already fixed with #704). Do you think you can merge it and release soon?.

kimrgrey:
Hi, @norman! What about this PR from @pabloh and new release? We are using `master` at the moment, but it looks like great idea to fix known issues and make new release.

PR #704, opened 2015-10-04
Fix sequentially_slugged bugs

pabloh:
As you suggested at #702, I tried the unreleased version with my projects and ran into some issues I think this PR should solve:

- Validating a record with sequential slugs would raise an error if the first candidate slug was nil.
- Using sequential slugs and candidate together arose some errors when calculating the sequential slug at some situations.

### DIFF
--- a/lib/friendly_id/sequentially_slugged.rb
+++ b/lib/friendly_id/sequentially_slugged.rb
@@ -26,7 +26,7 @@ module FriendlyId
       end
 
       def next_slug
-        slug + sequence_separator + next_sequence_number.to_s
+        [slug, next_sequence_number].compact.join(sequence_separator)
       end
 
     private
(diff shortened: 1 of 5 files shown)


lib/friendly_id/sequentially_slugged.rb
```diff
@@ -26,7 +26,7 @@ def initialize(scope, slug, slug_column, sequence_separator)
       end
 
       def next_slug
-        slug + sequence_separator + next_sequence_number.to_s
+        [slug, next_sequence_number].compact.join(sequence_separator)
```
kimrgrey: As for me, it's not good idea to run `next_sequence_number` when slug is `nil`. We can just return `nil` in this case to be sure that slug was not generated at all.

kimrgrey: I think it's better to fix this problem this way: https://github.com/norman/friendly_id/pull/720.

PR #706, opened 2015-10-15
Add a meaningful message to AR::RecordNotFound

elia:
That would greatly improve the debugging experience.


elia:
The failures seem unrelated

kimrgrey:
Well, may be it's good idea, personally I am not sure. Anyway, this PR was sent into `5-0-stable`, but this branch is staled. Please, use master instead.

PR #707, opened 2015-10-15
Update docs to use v5.0+ syntax for finding

cllns:
Since including `finders` isn't mentioned (and it's out of scope), we should include the 5.0+ syntax instead of the old one.


kimrgrey:
It is reasonable as for me because current examples just not works :+1:

kimrgrey:
@cllns, thank you for your contribution!

PR #718, opened 2016-01-01
Fixed problem that occurred when module :history was used after :finders

### LINKED ISSUE #717: finders + history are not working with relations
It looks like support of old-style finders in combination with history plugin is not working on relations. Please, review following example for details:

```ruby
begin
  require 'bundler/inline'
rescue LoadError => e
  $stderr.puts 'Bundler version 1.10 or later is required. Please update your Bundler'
  raise e
end

gemfile(true) do
  source 'https://rubygems.org'
  gem 'activerecord', '4.2.5'
  gem 'sqlite3'
  gem 'friendly_id', '~> 5.1.0', :require => false
end

require 'active_record'
require 'minitest/autorun'
require 'logger'
require 'friendly_id'

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :posts, force: true  do |t|
    t.string :title
    t.string :slug
    t.integer :category_id
  end

  create_table :categories, force: true do |t|
    t.string :name
  end

  create_table :friendly_id_slugs do |t|
    t.string   :slug,           :null => false
    t.integer  :sluggable_id,   :null => false
    t.string   :sluggable_type, :limit => 50
    t.string   :scope
    t.datetime :created_at
  end

  add_index :friendly_id_slugs, :sluggable_id
  add_index :friendly_id_slugs, [:slug, :sluggable_type], length: { slug: 140, sluggable_type: 50 }
  add_index :friendly_id_slugs, [:slug, :sluggable_type, :scope], length: { slug: 70, sluggable_type: 50, scope: 70 }, unique: true
  add_index :friendly_id_slugs, :sluggable_type
end

FriendlyId.defaults do |config|
  config.use :slugged
  config.use :finders
  config.use :history
end

class Category < ActiveRecord::Base
  has_many :posts
end

class Post < ActiveRecord::Base
  extend FriendlyId

  friendly_id :title

  belongs_to :category

  def should_generate_new_friendly_id?
    slug.blank? || title_changed?
  end
end

class BugTest < Minitest::Test
  def test_association_stuff
    category = Category.create(:name => "Test Category")
    post = category.posts.create!(:title => "Original Post")
    slug = post.slug
    post.update!(:title => "Changed Post")

    assert_equal post, Post.find(slug) # it works very well
    assert_equal post, category.posts.find(slug) # it doesn't work
  end
end
```

kimrgrey:
Fixed #717.

Module `:finders` includes methods from `friendly_id_config.finder_methods` into base class. But module `:history` replaces this methods. So, to make `:finders` and `:history` works together those methods must be reincluded into base class during initialization of `:history`.

### DIFF
--- a/gemfiles/Gemfile.rails-4.0.rb
+++ b/gemfiles/Gemfile.rails-4.0.rb
@@ -17,7 +17,7 @@ group :development, :test do
 
   platforms :ruby, :rbx do
     gem 'sqlite3'
-    gem 'mysql2'
+    gem 'mysql2', '~> 0.3.10'
     gem 'pg'
     gem 'redcarpet'
   end
--- a/gemfiles/Gemfile.rails-4.1.rb
+++ b/gemfiles/Gemfile.rails-4.1.rb
@@ -16,7 +16,7 @@ group :development, :test do
 
   platforms :ruby, :rbx do
     gem 'sqlite3'
-    gem 'mysql2'
+    gem 'mysql2', '~> 0.3.13'
     gem 'pg'
     gem 'redcarpet'
   end
--- a/lib/friendly_id/finders.rb
+++ b/lib/friendly_id/finders.rb
@@ -83,13 +83,10 @@ for models that use FriendlyId with something similar to the following:
 
       # Support for friendly finds on associations for Rails 4.0.1 and above.
       if ::ActiveRecord.const_defined?('AssociationRelation')
+        model_class.extend(ClassMethods)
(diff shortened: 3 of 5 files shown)


parndt:
looks good to me!

kimrgrey:
Well. Those changes looks good for 2 contributors. And this PR solves testing problem with `mysql` versions. Without this commit all other PR-s based on `master` will be broken. So, I think it's reasonable to merge it by myself.

PR #720, opened 2016-01-07
Frienldy_id should not generate a slug when all of canidates is nil.

kimrgrey:
Previously `friendly_id_config.base` was checked on presence in `should_generate_new_friendly_id?`. And let's imagine that we have multiple slug candidates, but all of them is nil. In this case we will have something like `[nil, nil]` as base for slug and `presence?` for this array will be evalutated to true. But all of slug canidates is nil, right? Booom! Exception during execution of `resolve_friendly_id_conflict`.

### DIFF
--- a/lib/friendly_id/sequentially_slugged.rb
+++ b/lib/friendly_id/sequentially_slugged.rb
@@ -4,13 +4,11 @@ module FriendlyId
       model_class.friendly_id_config.use :slugged
     end
 
-    def should_generate_new_friendly_id?
-      send(friendly_id_config.base).present? && super
-    end
-
     def resolve_friendly_id_conflict(candidate_slugs)
+      candidate = candidate_slugs.first
+      return if candidate.nil?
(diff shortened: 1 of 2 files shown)


parndt:
I don't think we should go removing methods which may be relied on like should_generate_new_friendly_id?

These should at least be deprecated first.  Someone may be relying on this functionality

Let me know if I'm missing something

parndt:
Actually, after a little more thought I can see where you're coming from and this looks okay to me.

kimrgrey:
Yep, but this method is still here: https://github.com/norman/friendly_id/blob/master/lib/friendly_id/slugged.rb#L298. It was just replaced by new implementation. And the main target of this replace is avoiding of slug generation when `base` is nil. But it not works when we have not just a one column, but more candidates.

kimrgrey:
@parndt, ok, cool! Thank you very much for a review! Please, take a short look on other PR-s marked as "looks good" when you will have time for it.

PR #722, opened 2016-01-07
Moved joins on slugs table from exists_by_friendly_id to scope_for_slug_generator

kimrgrey:
Call of `joins(:slugs)` in `FriendlyId::History#exists_by_friendly_id?` was expected by `FriendlyId::History#scope_for_slug_generator`, but it was useless for `exists_by_friendly_id?` itself and led to additional INNER JOIN in query without any reason. So, it was moved.

### DIFF
--- a/lib/friendly_id/history.rb
+++ b/lib/friendly_id/history.rb
@@ -79,7 +79,7 @@ method.
       include ::FriendlyId::FinderMethods
 
       def exists_by_friendly_id?(id)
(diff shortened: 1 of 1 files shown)


kimrgrey:
@norman, @parndt, what do you think about this PR? ^_^ It's not critical and everything works without this change. But I was surprised a little bit by this `join` when have been reading codebase of the project. And I spent some time to just find out why it's added into the scope but not used by method `exists_by_friendly_id?` itself.

parndt:
Fair enough :white_check_mark:

PR #724, opened 2016-01-12
Conditionally turn off :dependent => :destory on FriendlyId::Slugs 

vkmita:
Hey guys!

This adds the ability not to delete historic slugs when the sluggable class is destroyed.

When using acts_as_paranoid on a sluggable models, we aren't actually deleting the models when calling destroy. If we delete the slugs, we’re losing the history. We're also setting up a condition where a unique constraint ends up getting violated.

This is a revival of https://github.com/norman/friendly_id/pull/268

### DIFF
--- a/lib/friendly_id/base.rb
+++ b/lib/friendly_id/base.rb
@@ -187,9 +187,15 @@ often better and easier to use {FriendlyId::Slugged slugs}.
     #   allows an alternate configuration syntax, and conditional configuration
     #   logic.
     #
+    # @option options [Symbol,Boolean] :dependent Available when using `:history`.
+    #   Sets the value used for the slugged association's dependent option. Use
+    #   `false` if you do not want to dependently destroy the associated slugged
+    #   record. Defaults to `:destroy`.
+    #
     # @yieldparam config The model class's {FriendlyId::Configuration friendly_id_config}.
     def friendly_id(base = nil, options = {}, &block)
       yield friendly_id_config if block_given?
+      friendly_id_config.dependent = options.delete :dependent
       friendly_id_config.use options.delete :use
       friendly_id_config.send :set, base ? options.merge(:base => base) : options
       include Model
--- a/lib/friendly_id/configuration.rb
+++ b/lib/friendly_id/configuration.rb
@@ -18,6 +18,9 @@ module FriendlyId
     # The module to use for finders
     attr_accessor :finder_methods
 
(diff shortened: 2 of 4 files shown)


lib/friendly_id/base.rb
```diff
@@ -190,6 +190,7 @@ module Base
     # @yieldparam config The model class's {FriendlyId::Configuration friendly_id_config}.
     def friendly_id(base = nil, options = {}, &block)
       yield friendly_id_config if block_given?
+      friendly_id_config.options = options
```
kimrgrey: As for me, it's better to follow existing concept of explicitly specified configuration options and not to add whole hash. By the way, this config is available only for `history` module. So, we should clarify this fact in method's documentation above.

vkmita:
Thanks @kimrgrey does this work?

kimrgrey:
@vkmita, cool! Looks much better to me :+1: And Travis CI said that everything is ok. I think it's good idea to add some tests for another policies, isn't it? We should at least add test to check that `:destroy` is used by default.  

@parndt, what do you think about this PR? As for me it's useful and should be merged.

vkmita:
I added a test that checks that :destroy is default. What other policies would you want tested?

kimrgrey:
@vkmita, it's enough for me. Let's wait :+1: from another one contributor to merge it :wink:

parndt:
Thanks :+1: looks like @norman wanted this for a previous version of FriendlyId and this is a good implementation.

benortiz:
Hello, I'm trying to use this in one of my models. I was originally using version 5.1, but updated to 5.2 beta just in case.

When I add `:dependent => false` as it is done in the test to my model, I get an error that says that dependent is not defined (`NoMethodError - undefined method `dependent=' for #<#<Class:0x007f83b9667da0>:0x007f83b9667d50>:`.

my model has the line:
```
friendly_id :slug_candidates, use: [:slugged, :history], :dependent => false
```

There isn't a lot of documentation that I could find on this, so this is the best I could come up with. I've tried various rearrangements of this to no avail. Perhaps I need to do something in the friendly_id initializer?

If there's any further information I can provide, let me know.

Thanks in advance for your help.

benortiz:
Ah, looks like it may have been my misunderstanding. I checked the beta and found out it wasn't included in that. I switched to master and everything worked as described. Thanks so much for an awesome gem!

PR #735, opened 2016-03-04
Configurable Route Generation

sjaveed:
* friendly_id now takes a routes option which, when set to anything but :friendlyid will generate standard rails style routes when used with e.g. polymorphic_path or form_for etc
* when the routes option isn't specified, it defaults to the current behavior

Given a model:
```ruby
class Team < ActiveRecord::Base
end
```
it generates:
```ruby
team_path(team)  # returns /team/123-awesome-team-name
team_path(team.id) # returns /team/123
polymorphic_path(team) # returns /team/123-awesome-team-name
```

With this patch, given a model like:
```ruby
class Team < ActiveRecord::Base
  friendly_id :name, use: :slugged, routes: :default
end
```
it generates:
```ruby
team_path(team) # should return /team/123
team_path(team.friendly_id) # should return /team/123-awesome-team-name
polymorphic_path(team) # should return /team/123
```


sjaveed:
@parndt here you go!

lib/friendly_id/base.rb
```diff
@@ -196,6 +196,7 @@ module Base
     def friendly_id(base = nil, options = {}, &block)
       yield friendly_id_config if block_given?
       friendly_id_config.dependent = options.delete :dependent
+      # friendly_id_config.routes = (options.delete(:routes) || :friendly)
```
parndt: I'm a little confused about this line of code and why it's commented out? :smile:

parndt:
@sjaveed that was lightning :zap: fast.  

Could we please have some tests to ensure that we don't regress this behaviour, and a note in the changelog file?

lib/friendly_id/base.rb
sjaveed: Good catch :-)  I think I'd added it in there before I realized I could leverage the set.  Commented it out to test the set and then didn't delete it before committing :-)  I'll remove this in the next commit.

sjaveed:
@parndt heh thanks :-)  I did say I was frustrated :-)  Let me see how to test this sucker properly.

sjaveed:
@parndt updated to fix an issue with the default value of routes and to add tests.  Any specific format you'd like me to follow for the Changelog?

parndt:
For changelog just something simple that links to this PR like in https://github.com/norman/friendly_id/blob/master/Changelog.md#520-not-released-yet

sjaveed:
@parndt updated the change log as well.  Do let me know if you need me to change anything else :-)

parndt:
Thanks, I merged it as 8531cdc :smile:

sjaveed:
Awesome!  Thanks @parndt!

PR #751, opened 2016-05-14
Fix SequentiallySlugged when used with History

kylefleming:
I encountered a crash when using `:sequentially-slugged` with `:history`. Since it does a join with the history table, it causes an ambiguous column name error:

```
ActiveRecord::StatementInvalid: SQLite3::SQLException: ambiguous column name: slug: SELECT "slug" FROM "articles" INNER JOIN "friendly_id_slugs" ON "friendly_id_slugs"."sluggable_id" = "articles"."id" AND "friendly_id_slugs"."sluggable_type" = ? WHERE ("articles"."id" != 3) AND (sluggable_id <> 3) AND ("slug" = 'test-name' OR "slug" LIKE 'test-name-%' ESCAPE '\')  ORDER BY LENGTH("slug") ASC, "slug" ASC
```

This fix adds disambiguation to `:sequentially-slugged` by including the table name whereever it uses a column name.

### DIFF
--- a/lib/friendly_id/sequentially_slugged.rb
+++ b/lib/friendly_id/sequentially_slugged.rb
@@ -10,16 +10,18 @@ module FriendlyId
       SequentialSlugCalculator.new(scope_for_slug_generator,
                                   candidate,
                                   friendly_id_config.slug_column,
-                                  friendly_id_config.sequence_separator).next_slug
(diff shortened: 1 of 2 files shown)


coneybeare:
Please merge into master

Issue #749, opened 2016-05-13
please support rails5  thanks !


FingersKey:

suramai:
so how to get this working with rails 5?

parndt:
Closing due to #753

PR #753, opened 2016-05-23
Reduce restriction on activerecord version

djsegal:
Start resolving #749


djsegal:
@nburkley do you think activerecord 5.0rc will break anything?

nburkley:
Should be fine. I'm running 5.0rc1 since it's release. I can update the
specs tomorrow.

parndt:
Doesn't this make it more restrictive?

PR #758, opened 2016-06-09
Ability to change slugs manually

sergio1990:
**What have done:**
1. Implemented processing use case when slug can be changed manually. For example, by content manager in the admin panel.
2. Added specs for new use case.

**The main point of this pull request**

By additional possibility for configuration (I mean override `#should_manually_change_friendly_id?` and `#should_generate_new_friendly_id?`) you can setup the next flow of changing slugs:
1. When sluggable object is created a slug is automatically generated based on a some field value.
2. When source field for the slug is changed the slug stays same.
3. In order to change slug there is an admin panel and content manager is able to change it. That new slug is checked with current sluggable objects slugs and with the history slugs, and can be prepended with the counter if this is needed. Eventually the final version of the slug also is added to the history.

The described above flow I was asked to implement in the one of your projects. Maybe it will be useful for someone.

### DIFF
--- a/lib/friendly_id/sequentially_slugged.rb
+++ b/lib/friendly_id/sequentially_slugged.rb
@@ -7,20 +7,21 @@ module FriendlyId
     def resolve_friendly_id_conflict(candidate_slugs)
       candidate = candidate_slugs.first
       return if candidate.nil?
-      column_table = self.class.table_name
       SequentialSlugCalculator.new(scope_for_slug_generator,
                                   candidate,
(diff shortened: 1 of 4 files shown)


parndt:
@norman seems like a decent feature to me

oyeanuj:
@sergio1990 @parndt Is this good to merge?

parndt:
Ah, sorry, with time there are now conflicts that need to be resolved. I won't be able to get to resolving those today with other priorities, so if anyone's keen to help that'd be wonderful.

PR #770, opened 2016-09-13
Require active record before using it

ryansch:
Trying to require friendly_id outside of rails results in `uninitialized constant FriendlyId::ActiveRecord`.  My use case is preloading gems with sidekiq swarm.

Adding a require fixes this use case.


parndt:
Thank you!

PR #773, opened 2016-09-29
Split Candidates#each in few methods

novikserg:
Hey,

This PR splits `Candidates#each` method, thus separating actual yielding from constructing/filtering elements. 

A minor change, but we can further improve from here: one idea is to create a `Candidate` class, and move slug construction/validation there. 

Should also solve CodeClimate Issue (https://codeclimate.com/github/norman/friendly_id/FriendlyId::Candidates#complexity)

Main idea is to make this part of code more readable (while keeping simplicity/flexibility)

### DIFF
--- a/lib/friendly_id/candidates.rb
+++ b/lib/friendly_id/candidates.rb
@@ -10,26 +10,35 @@ module FriendlyId
 
     def initialize(object, *array)
       @object = object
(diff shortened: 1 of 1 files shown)


parndt:
Thanks @novikserg

PR #774, opened 2016-10-06
Sequentially slugged bug fixes

vddgil:
Add fix when used which History thanks to @kylefleming

Use last candidate slug in place of first


parndt:
Thanks @vddgil

PR #779, opened 2016-12-02
Assign to configured slug column, not "slug", when validation fails

### LINKED ISSUE #765: undefined method `slug=' for Model
I am using master branch with rails 5 there is no slug field in my db as i am using another field using 
`slug_column: :column_name`

after validations it is raising `undefined method 'slug=' *` error
This error is caused by `self.slug = diff.first` from `lib/friendly_id/slugged.rb:332` in `unset_slug_if_invalid`
I changed this to `slug = diff.first` by removing `self.` and that solved the issue

cgunther:
1f8af74b mistakenly assigns to the literal `slug` method, rather than the configured `slug_column`. As a result, models using a custom `slug_column` get an error raised when validations fail that `slug=` is an undefined method.

Fixes #765.

### DIFF
--- a/lib/friendly_id/slugged.rb
+++ b/lib/friendly_id/slugged.rb
@@ -327,9 +327,9 @@ Github issue](https://github.com/norman/friendly_id/issues/185) for discussion.
     private :slug_generator
 
     def unset_slug_if_invalid
-      if errors.present? && attribute_changed?(friendly_id_config.query_field)
+      if errors.present? && attribute_changed?(friendly_id_config.query_field.to_s)
         diff = changes[friendly_id_config.query_field]
-        self.slug = diff.first
+        send "#{friendly_id_config.slug_column}=", diff.first
       end
     end
     private :unset_slug_if_invalid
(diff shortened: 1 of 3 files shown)


marshall-lee:
@parndt @kimrgrey @norman 

Please please merge this one and release it in "5.2.1"! :)

We really need this `unset_slug_if_invalid` callback and other changes introduced in "5.2.0" but our tests fail on one of the models with non-standard slug field.

ezekg:
Spree Commerce is also waiting on this patch: https://github.com/spree/spree/issues/7756.

ezekg:
Is there anyway we could get this merged so that we can update friendly_id for Spree 3.3?

parndt:
Thanks @cgunther ❤️

vfonic:
+1 thanks for all the work! /cc @parndt

parndt:
@vfonic done 😄 thanks for the reminder.

vfonic:
Thanks! 🍾 🎉 😄

marshall-lee:
Thanks!

PR #782, opened 2016-12-07
Change ActiveRecord::Base to ApplicationRecord

mrjlynch:
Rails 5 Active Record models now inherit from `ApplicationRecord` by default. I thought it appropriate to reflect this in the readme.


parndt:
@mrjlynch thanks, that probably does make sense now.

PR #788, opened 2017-01-30
Predictive Match for `id`, Fix #785

### LINKED ISSUE #785: Predictive Match and Case Insensitivity

ghbooth12:
Hello. This PR is about fixing the second issue of #785 by printing the similar ids with the `RecordNotFound` error message.

The current `RecordNotFound` error message is:
```bash
ActiveRecord::RecordNotFound (can't find record with friendly id: "john"):
```
The new `RecordNotFound` error message with predictive match for `id` is:
```bash
ActiveRecord::RecordNotFound (can't find record with friendly id: "john"
Do you mean? "john-doe", "john-smith"):
```

### DIFF
--- a/lib/friendly_id/finder_methods.rb
+++ b/lib/friendly_id/finder_methods.rb
@@ -33,7 +33,7 @@ module FriendlyId
     # `find`.
     # @raise ActiveRecord::RecordNotFound
     def find_by_friendly_id(id)
-      first_by_friendly_id(id) or raise ActiveRecord::RecordNotFound, "can't find record with friendly id: #{id.inspect}"
+      first_by_friendly_id(id) or raise ActiveRecord::RecordNotFound, "can't find record with friendly id: #{id.inspect}#{similar_id(id)}"
     end
 
     def exists_by_friendly_id?(id)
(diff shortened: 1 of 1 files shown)


parndt:
This could be helpful, if it were made configurable. I think in many circumstances we wouldn't want to leak possible other URLs

parndt:
@ghbooth12 @StarWar are you still interested in this feature request?

StarWar:
@parndt yes. It has to be configurable. It'll be even better if paths and even subpaths can be predicted in case of a record not found... currently, I'm using a custom method in one of my popular blog website where people have linked to the posts with all sort of URLs (stripped a portion of the url, upcased a few words or all caps and all possible variants) and in order to not lose the linking and traffic... it'll be very good to have a configurable option as to which models to search for and do a case-insensitive search across models... currently, I'm utilizing elastic search to search across models in case of 404 and redirecting to the first result.

PR #802, opened 2017-04-07
Allow to set size limit for slug

avokhmin:
By: https://github.com/norman/friendly_id/issues/190

Issue #190, opened 2011-11-24
best way to limit slug length?


ghost:
I managed to save an item that has an epically long slug, but now I can't perform CRUD on it via a browser due to ...

```
Request-URI Too Large

WEBrick::HTTPStatus::RequestURITooLarge
WEBrick/1.3.1 (Ruby/1.9.2/2011-07-09) at macbook.local:3000
```

What would be the best way to limit the slug's length. Why are they called slugs anyway, yuck.

Dieglock:
+ another 1 to @Sjors. Was very handy...

Empact:
@norman I think it was adopted by the newspaper industry based on its use in the typesetting industry: https://en.wikipedia.org/wiki/Slug_(typesetting)

PR #809, opened 2017-05-16
Allow to set size limit for slug

avokhmin:
By: 

- https://github.com/norman/friendly_id/issues/190
- https://github.com/norman/friendly_id/pull/802

### DIFF
--- a/lib/friendly_id/slugged.rb
+++ b/lib/friendly_id/slugged.rb
@@ -301,16 +301,55 @@ Github issue](https://github.com/norman/friendly_id/issues/185) for discussion.
       send(friendly_id_config.slug_column).nil? && !send(friendly_id_config.base).nil?
     end
 
+    # Public: Resolve conflicts.
+    #
+    # This method adds UUID to first candidate and truncates (if `slug_limit` is set).
+    #
+    # Examples:
+    #
+    #   resolve_friendly_id_conflict(['12345'])
+    #   # => '12345-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'
+    #
+    #   FriendlyId.defaults { |config| config.slug_limit = 40 }
+    #   resolve_friendly_id_conflict(['12345'])
+    #   # => '123-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'
+    #
+    # candidates - the Array with candidates.
+    #
+    # Returns the String with new slug.
     def resolve_friendly_id_conflict(candidates)
-      candidate = candidates.first
-      uid = SecureRandom.uuid
-      if candidate && friendly_id_config.slug_limit
-        max_candidate_size = friendly_id_config.slug_limit - uid.size - friendly_id_config.sequence_separator.size
-        max_candidate_size = [max_candidate_size, 0].max
-        candidate = candidate[0...max_candidate_size]
-      end
-      [candidate, uid].compact.join(friendly_id_config.sequence_separator)
+      uuid = SecureRandom.uuid
+      [
+        apply_slug_limit(candidates.first, uuid),
+        uuid
+      ].compact.join(friendly_id_config.sequence_separator)
     end
+
+    # Private: Apply slug limit to candidate.
+    #
+    # candidate - the String with candidate.
+    # uuid      - the String with UUID.
+    #
+    # Return the String with truncated candidate.
+    def apply_slug_limit(candidate, uuid)
+      return candidate unless candidate && friendly_id_config.slug_limit
+
+      candidate[0...candidate_limit(uuid)]
+    end
+    private :apply_slug_limit
+
+    # Private: Get max length of candidate.
+    #
+    # uuid - the String with UUID.
+    #
+    # Returns the Integer with max length.
+    def candidate_limit(uuid)
+      [
+        friendly_id_config.slug_limit - uuid.size - friendly_id_config.sequence_separator.size,
+        0
+      ].max
+    end
+    private :candidate_limit
 
     # Sets the slug.
     def set_slug(normalized_slug = nil)
--- a/lib/friendly_id/slugged.rb
+++ b/lib/friendly_id/slugged.rb
@@ -288,7 +288,9 @@ Github issue](https://github.com/norman/friendly_id/issues/185) for discussion.
     # @param [#to_s] value The value used as the basis of the slug.
     # @return The candidate slug text, without a sequence.
     def normalize_friendly_id(value)
-      value.to_s.parameterize
+      value = value.to_s.parameterize
+      value = value[0...friendly_id_config.slug_limit] if friendly_id_config.slug_limit
+      value
     end
 
     # Whether to generate a new slug.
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -2,10 +2,10 @@ language: ruby
 cache: bundler
 
 rvm:
+  - 2.4.0
   - 2.3.0
   - 2.2.0
   - 2.1.0
-  - 2.0.0
```
avokhmin: I think that we can do it, because [Support of Ruby 2.1 has ended](https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/)

.travis.yml
```diff
@@ -19,19 +19,15 @@ gemfile:
...
-    - rvm: 2.0.0
-      gemfile: gemfiles/Gemfile.rails-5.0.rb
     - rvm: 2.1.0
       gemfile: gemfiles/Gemfile.rails-5.0.rb
     - rvm: 2.2.0
       gemfile: gemfiles/Gemfile.rails-5.0.rb
+    - rvm: 2.4.0
+      gemfile: gemfiles/Gemfile.rails-4.0.rb
+    - rvm: 2.4.0
+      gemfile: gemfiles/Gemfile.rails-4.1.rb
```
avokhmin: https://travis-ci.org/Shuttlerock/friendly_id/jobs/232872362

```
TypeError: Cannot visit Integer
    /home/travis/build/Shuttlerock/friendly_id/gemfiles/vendor/bundle/ruby/2.4.0/gems/arel-5.0.1.20140414130214/lib/arel/visitors/visitor.rb:28:in `rescue in visit'
```

http://weblog.rubyonrails.org/2017/3/20/Rails-5-1-rc1/

> As per our maintenance policy, the release of Rails 5.1 will mean that bug fixes will only apply to 5.1.x, regular security issues to 5.1.x and 5.0.x, and severe security issues to 5.1.x, 5.0.x, and 4.2.x. This means 4.x and below will essentially be unsupported!

I think that we can skip `4.0` and `4.1` here too.

avokhmin:
@norman , can you write a feedback?

norman:
Thanks for your time and efforts, @avokhmin! I'm not currently actively maintaining FriendlyId any more - @parndt is the main maintainer now, but I do have some feedback.

First, the changes to the CI setup look great, but it would probably be better to have those done in a separate PR from this in order to keep things focused. I suspect those changes could me merged right away, whereas the other stuff might need some more time and discussion. Thank you for that! Keeping the CI setup working properly is useful and valuable work.

Regarding the slug limit, this is a feature that used to be in FriendlyId 3.0, but that I removed from FriendlyId 4.0. My idea at the time was that the size limit could be controlled with Active Record validations or column size limits, so it didn't need to be in the library. Version 3.0 had grown to be so large that I wanted to remove as much as possible from the library to keep it easier to maintain and less likely to conflict with other libraries.

The natural lifecycle of plugins like this is to add, add, add and add more and more and more features; everybody has a good reason for wanting to add the thing they want, and it might be legitimately useful and be well written (as your code looks to be here). But if we accept every one that seems justifiable and well written, eventually the library becomes "bloated". Then, somebody else comes along and writes a competing library that's "like FriendlyId but with less bloat", repeat the cycle over and over again. I think we've all seen this. This is what I wanted to avoid in FriendlyId 4.0 and so I rewrote it from the ground up to be simpler and smaller, and less likely to be a library that projects would want to remove because of bloat. It may sound counterintuitive but I've always encouraged people to use FriendlyId for as little as possible, which I think makes it more useful over the long term.

Your feature looks legitimately useful and well coded, so I'll leave it up to @parndt and other folks if they want to include it, but that's the background on why this feature is not currently included in the library.

parndt:
Thank you @norman 

@avokhmin I agree with @norman here, can you introduce 9ff5fa6fae7c0319536c77060a5da405269b6056 as its own PR so that it's easier to reason about this PR?

Let me know if you would like pointers about that and I'll do my best to help.

avokhmin:
> @avokhmin I agree with @norman here, can you introduce 9ff5fa6 as its own PR so that it's easier to reason about this PR?

@parndt done, see: https://github.com/norman/friendly_id/pull/824

avokhmin:
`git rebase master`

@parndt, PR updated.

lib/friendly_id/slugged.rb
```diff
@@ -300,7 +302,14 @@ def should_generate_new_friendly_id?
...
     def resolve_friendly_id_conflict(candidates)
-      [candidates.first, SecureRandom.uuid].compact.join(friendly_id_config.sequence_separator)
+      candidate = candidates.first
+      uid = SecureRandom.uuid
+      if candidate && friendly_id_config.slug_limit
+        max_candidate_size = friendly_id_config.slug_limit - uid.size - friendly_id_config.sequence_separator.size
+        max_candidate_size = [max_candidate_size, 0].max
+        candidate = candidate[0...max_candidate_size]
+      end
+      [candidate, uid].compact.join(friendly_id_config.sequence_separator)
```
parndt: OK just one last comment. I would use the 'extract method' refactor here:

```ruby
def resolve_friendly_id_conflict(candidates)
  uuid = SecureRandom.uuid
  [
    apply_slug_limit(candidates.first, uuid), 
    uuid
  ].compact.join(friendly_id_config.sequence_separator)
end

def apply_slug_limit(candidate, uuid)
  return candidate unless candidate && friendly_id_config.slug_limit
  
  candidate[0...slug_limit(uuid)]
end
private :apply_slug_limit

def slug_limit(uuid)
  [
    friendly_id_config.slug_limit - uuid.size - friendly_id_config.sequence_separator.size,
    0
  ].max
end
private :max_candidate_size_for_uuid
```

I haven't tried the code, but would something along those lines work?
My rationale is that it just makes the original method easier to reason about rather than having to read the code a few times to figure out what's going on with the conditional.  It also means others can use this functionality without rewriting it if they override `resolve_friendly_id_conflict`. 😄

parndt: There may even be a simpler version waiting to be found, but I wanted to communicate my thoughts with Ruby code. 👍

avokhmin: I like it, looks much better :+1:

avokhmin:
@parndt, PR updated.

avokhmin:
@parndt , something wrong here?

parndt:
No, just a lack of time at the computer 😄 thanks a lot @avokhmin

avokhmin:
@parndt , can I create a PR for next release? Or when it can be?

parndt:
@avokhmin please, create a PR 😄 👍  thanks

avokhmin:
@parndt , done, see: https://github.com/norman/friendly_id/pull/828

nuzelac:
@avokhmin 

Does setting `slug_limit` in the configuration work for you? 
I'm using Rails 5.0.6 and friendly_id 5.2.3. I can use `:slug_limit` from the model:
```ruby
friendly_id :name, use: :slugged, slug_limit: 50
```

But if I move it into the initializer:
```ruby
#
# By default, slug has no size limit, but you can change it if you wish.
#
config.slug_limit = 50
```
I receive this error after running `Product.find_each(&:save)`:
```ruby
NoMethodError: undefined method `slug_limit=' for #<#<Class:0x007fd3ea4cf4a8>:0x007fd3ea4cf458>
```

avokhmin:
@nuzelac , I use it only in the `config/initializers/friendly_id.rb`:

```
FriendlyId.defaults do |config|
  config.use :slugged, :reserved
  config.slug_limit = 255
end

Gemfile.lock

    friendly_id (5.2.2)
    rails (5.1.4)
```

do you use `:slugged `?

nuzelac:
Thank you @avokhmin. I added `config.use :slugged` and now it works from the initializer.

PR #821, opened 2017-07-26
Make sluggable unscoped by default

joostvanrijn:


parndt:
Thank you - I've merged #838 which also fixes this in the same way but provides tests for it. 😄

PR #829, opened 2017-09-16
Remove extra ``` in README

lowderl:
There was an extra set of backticks in the README, so **```shell** was showing up in the code block.


parndt:
Thanks! 😄

PR #831, opened 2017-09-21
option to treat reserved words as conflicts, fix #830

### LINKED ISSUE #830: Regenerating slug when same as reserved words
Hi @parndt, thank you for supporting this library! Not sure if this question is a bug or a feature request but would really appreciate your thoughts.

**TLDR;** *How does one deal with (ideally, globally) regenerating slugs or providing slug candidates when the initial slug choice may conflict with a reserved word?*

---
**Detailed version**
In my app, there are cases when I get this message:
```sh
ActiveRecord::RecordInvalid: Validation failed: Friendly is reserved
```

Model:
```rb
class Tag < ApplicationRecord
  extend FriendlyId
  friendly_id 	:slug_candidates,
	use: 		[:slugged, :history]
  
  # ... more model code ...
  
  def slug_candidates
    [:name]
  end
end
```

Now, based on the documentation:
> Now that candidates have been added, FriendlyId no longer uses a numeric sequence to differentiate conflicting slug, but rather a UUID (e.g. something like 2bc08962-b3dd-4f29-b2e6-244710c86106). This makes the codebase simpler and more reliable when running concurrently, at the expense of uglier ids being generated when there are conflicts.

..I'd expect to see a UUID when the candidates conflict with the reserved words. So, is the error being thrown a bug or expected behavior?

And in either case, what is (an ideally global) workaround that you'd recommend?

Thank you!

oehlschl:
Closes #830 

This adds an option to FriendlyId::Reserved::Configuration to treat reserved words as conflicts rather than exceptions. When there is no good candidate, a UUID will be appended, matching the existing conflict behavior.

This could be implemented differently, but I tried to keep with the same patterns used in the gem.

Testing: 
-  Changes are opt-in, so behavior is non-breaking.
- New and old specs pass.

### DIFF
--- a/lib/friendly_id/reserved.rb
+++ b/lib/friendly_id/reserved.rb
@@ -46,6 +46,7 @@ message to a different field. For example:
     # {FriendlyId::Configuration FriendlyId::Configuration}.
     module Configuration
       attr_accessor :reserved_words
+      attr_accessor :treat_reserved_as_conflict
     end
   end
 end
--- a/lib/friendly_id/scoped.rb
+++ b/lib/friendly_id/scoped.rb
@@ -132,7 +132,7 @@ an example of one way to set this up:
     private :scope_for_slug_generator
 
     def slug_generator
-      friendly_id_config.slug_generator_class.new(scope_for_slug_generator)
+      friendly_id_config.slug_generator_class.new(scope_for_slug_generator, friendly_id_config)
     end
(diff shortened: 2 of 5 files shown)


parndt:
Very nice! I'll wait for CI to finish its run.

oehlschl:
Awesome, thanks!

PR #832, opened 2017-09-28
migration version for rails 5

AnatolyShirykalov:


lib/friendly_id/migration.rb
```diff
@@ -1,4 +1,4 @@
-class CreateFriendlyIdSlugs < ActiveRecord::Migration
+class CreateFriendlyIdSlugs < ActiveRecord::Migration<%= migration_version %>
```
parndt: I think that this interpolation won't work as this is a ruby file. I think it would need to be renamed to `migration.rb.erb` so that it could be parsed as erb

AnatolyShirykalov: It works. For example Devise use it without «erb» prefix.

https://github.com/plataformatec/devise/blob/master/lib/generators/active_record/templates/migration.rb

parndt: I believe it is the reason for the syntax error in the builds: https://travis-ci.org/norman/friendly_id/jobs/280775757#L634

bananaappletw: I guess it's ruby' version problem.
I just check devise project.
Almost the same scenario, but devise pass the travis ci.
I am trying to use ruby 2.4.1 instead of 2.4.0.

bananaappletw: I just test gem locally.
This patch is correct.
But the test system is broken.
Rake task is not test functionally.

parndt: Devise's version works because it's a template file which all use erb tags. This file is directly required by another Ruby file which means when we run tests we get this output:

```
ruby/2.4.2/gems/activesupport-4.2.10/lib/active_support/dependencies.rb:274:in `require': 
friendly_id/lib/friendly_id/migration.rb:14: syntax error, unexpected keyword_end, expecting end-of-input (SyntaxError)
  end
```

If we moved to Rails 4.2 being the minimum supported version, this could be similarly solved by just generating migrations with version `4.2`:

```ruby
class CreateFriendlyIdSlugs < ActiveRecord::Migration[4.2]
```

It might be worth it to not support Rails 4.0 and 4.1 as the Rails team don't support them anymore either..

What do you think?

bananaappletw: @parndt 
I agree with you.
It could be a quick fix for now.
But if we have time, we still need to rewrite the test system running for generating the migration template.
Then if anyone is willing to rewrite it, we could change back to this pull request.
@AnatolyShirykalov

zhisme:
This one also can be closed.
Fixed here [PR-849](https://github.com/norman/friendly_id/pull/849)

PR #835, opened 2017-09-30
Removes multiple warnings when running 'bundle exec rake test'

padi:
1. ActiveSupport deprecation warning for ActiveSupport 4.2:

    DEPRECATION WARNING: You did not specify a value for the configuration option `active_support.test_order`. In Rails 5, the default value of this option will change from `:sorted` to `:random`.
    To disable this warning and keep the current behavior, you can add the following line to your `config/environments/test.rb`:

      Rails.application.configure do
        config.active_support.test_order = :sorted
      end

    Alternatively, you can opt into the future behavior by setting this option to `:random`. (called from test_order at …/gems/activesupport-4.2.10/lib/active_support/test_case.rb:42)

2. Ruby warning:

    friendly_id/test/history_test.rb:251: warning: assigned but unused variable - e

3. Ruby warning:

    friendly_id/lib/friendly_id/finder_methods.rb:53: warning: ambiguous first argument; put parentheses or a space even after `/' operator

### DIFF
--- a/lib/friendly_id/finder_methods.rb
+++ b/lib/friendly_id/finder_methods.rb
@@ -50,7 +50,7 @@ module FriendlyId
       when :integer
         Integer(id, 10) rescue false
       when :uuid
-        id.match /\A[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\z/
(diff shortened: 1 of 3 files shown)


parndt:
Hey, thanks for this. On some versions that we test against this causes an error: https://travis-ci.org/norman/friendly_id/jobs/281625073

padi:
@parndt looks like it's passing now except for one build that seems to be stuck: 487.9

https://travis-ci.org/norman/friendly_id/builds/282732285?utm_source=github_status&utm_medium=notification

parndt:
Thank you so much! 🎉

PR #838, opened 2017-10-04
Slugs should always find their sluggable, even when they are soft deleted by acts_as_paranoid

padi:
Probably fixes #822, as the test case is likely what he has encountered. His report most probably required 2 things to be true:

1. He used `friendly_id ... :dependent => false`
2. friendly_id is used in a Rails 5.0 app

Previously, when a sluggable is soft deleted and a slug persists
(i.e. :dependent => false), slug can't find its corresponding
sluggable.

Additionally, in Rails 5.0, belongs_to_required_by_default has been
introduced to be true by default (unless changed back to false), therefore making Friendly::Slug
invalid with the error message 'Sluggable must exist'.

PR #839, opened 2017-10-06
Removes redundant HistoryTest

padi:
DependentDestroyTest used to inherit from HistoryTest,
which run tests for a DependentDestroyTest#model_class
that was not overriden anyway.

Previously had 431 tests. Now we have 403 total tests

### DIFF
--- a/test/history_test.rb
+++ b/test/history_test.rb
@@ -172,7 +172,7 @@ class HistoryTestWithAutomaticSlugRegeneration < HistoryTest
   end
 end
 
-class DependentDestroyTest < HistoryTest
+class DependentDestroyTest < TestCaseClass
 
   include FriendlyId::Test
 


padi:
Since this is only a one-liner that's already applied to other parts of the tests, and since https://github.com/norman/friendly_id/pull/838 is failing for the same builds, I'm lead to believe that some jruby builds are currently failing regardless of the pull request.

padi:
@parndt the DependentDestroyTest tests against 2 classes (FalseManual, DefaultManual) that use the history module with the `:dependent` option set to `false` or `true` (default), but it doesn't make use of the tests in HistoryTest class by defining a new DependentDestroyTest#model_class.

parndt:
Thanks @padi

PR #852, opened 2018-01-11
Defer loading of ActiveRecord to avoid config issues

### LINKED ISSUE #823: Friendly ID can break configuration via initializers
When upgrading to Rails 5.0, the `rails app:upgrade` command will generate `config/initializers/new_framework_defaults.rb`, which contains the following lines:

```ruby
# Require `belongs_to` associations by default. Previous versions had false.
Rails.application.config.active_record.belongs_to_required_by_default = true
```

If Friendly ID is present, this setting will not actually have any effect in the app; this can be verified in the console:

```ruby
ActiveRecord::Base.belongs_to_required_by_default => nil
```

I have replicated it in a brand new Rails 5.1 app.

In a nutshell, because Friend ID references `ActiveRecord::Base` while it is loading (see [object_utils.rb](https://github.com/norman/friendly_id/blob/master/lib/friendly_id/object_utils.rb#L69)), it triggers the `on_load` event (see [active_record/railtie.rb](https://github.com/rails/rails/blob/v5.1.2/activerecord/lib/active_record/railtie.rb#L110) in Rails itself), which transmutes all the existing config settings into `ActiveRecord::Base`. 

Rails will only call the `on_load` event once, and because it happens before any initializers have been read, this means that any configuration of ActiveRecord in an initializer that takes place via the `Rails.application.config.active_record` object is essentially ignored when Friendly ID is present. This impacts Rails 5.0 and Rails 5.1 at the moment.

### Possible solutions

I can think of two basic ways to work around this:

#### Defer loading Friendly ID

First, change `Gemfile` to include `gem 'friendly_id', require: false`. Then, add to `config/initializers/friendly_id`:

```ruby
initialize do
  require 'friendly_id'
end
```

#### Defer loading ActiveRecord

First move the `include` call and references to `ActiveRecord::Base` into a method, something like `FriendlyID.setup`:

```ruby
module FriendlyId
  def self.setup
    [
      ActiveRecord::Base,
      Array,
      FalseClass,
      Hash,
      NilClass,
      Numeric,
      Symbol,
      TrueClass
    ].each { |klass| klass.send(:include, FriendlyId::UnfriendlyUtils) }
  end
end
```

That's just a rough sketch, but the key is that there are no references to the `ActiveRecord` constant that get hit merely by requiring these files. Then, either add a call `FriendlyID.setup` in an `initializer` block as above, or add a `Railtie` for FriendlyID:

```ruby
module FriendlyId
  class Railtie < Rails::Railtie
    initializer "friendly_id.setup" do
      ActiveSupport.on_load(:active_record) do
        FriendlyId.setup
      end
    end
  end
end

# in lib/friendly_id.rb:
require 'friendly_id/railtie' if defined?(Rails)
```

I'd be very happy to work up a pull request for either approach; in the short term I'll be using the first one in our own app as a workaround.

lazyatom:
If FriendlyId references `ActiveRecord::Base` while it is loading as part of a Rails application, it triggers the `on_load` event for ActiveRecord, which as part of the loading process sets the current configuration values of ActiveRecord. If any initializers try to influence that configuration (e.g. changing defaults in `new_framework_defaults.rb` for an app that's been upgraded), these changes will have no effect because the configuration has already been set in place.

To avoid this, we need to avoid triggering the `on_load` event by wrapping the reference to `ActiveRecord::Base` in an initializer block. The simplest way of doing this is with a `Railtie`, which will be automatically loaded if FriendlyId is required as part of Rails.

Testing this change is unfortunately quite hard; it would require setting up a Rails test harness application to exercise the Railtie. Since FriendlyId doesn't currently have that, I've tried to keep the
change small.

Fixes #823

### DIFF
--- a/lib/friendly_id.rb
+++ b/lib/friendly_id.rb
@@ -4,6 +4,7 @@ require "friendly_id/base"
 require "friendly_id/object_utils"
 require "friendly_id/configuration"
 require "friendly_id/finder_methods"
+require 'friendly_id/railtie' if defined?(Rails)
 
 =begin
 
--- a/lib/friendly_id/object_utils.rb
+++ b/lib/friendly_id/object_utils.rb
@@ -2,7 +2,6 @@ module FriendlyId
   # Instances of these classes will never be considered a friendly id.
   # @see FriendlyId::ObjectUtils#friendly_id
   UNFRIENDLY_CLASSES = [
-    ActiveRecord::Base,
     Array,
     FalseClass,
     Hash,
@@ -59,6 +58,10 @@ module FriendlyId
       true
     end
   end
+
+  def self.mark_as_unfriendly(klass)
+    klass.send(:include, FriendlyId::UnfriendlyUtils)
+  end
 end
 
 Object.send :include, FriendlyId::ObjectUtils
@@ -66,4 +69,4 @@ Object.send :include, FriendlyId::ObjectUtils
 # Considered unfriendly if object is an instance of an unfriendly class or
 # one of its descendants.
 
-FriendlyId::UNFRIENDLY_CLASSES.each { |klass| klass.send(:include, FriendlyId::UnfriendlyUtils) }
+FriendlyId::UNFRIENDLY_CLASSES.each { |klass| FriendlyId.mark_as_unfriendly(klass) }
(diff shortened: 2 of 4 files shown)


parndt:
Thank you!

PR #853, opened 2018-01-22
Refactor History::FinderMethods: use base implementations.

marshall-lee:
It also makes `first_by_friendly_id` better because base implementation
in `FriendlyId::FinderMethods` uses `find_by` which is optimized in
Rails 4.2 (so called "Adequate Record").


parndt:
This makes total sense, thank you!

PR #857, opened 2018-03-01
Remove second 'private' statement from  `lib/friendly_id/candidates.rb`

shiroginne:
Hello,

it's nothing critical, just a minor style PR: `lib/friendly_id/candidates.rb` had 2 **private** statements 😊


parndt:
Thank you!

PR #858, opened 2018-03-02
Added migration version if the project is using rails 5.x

an-nasir:
Hi, 
please review my request.

### DIFF
--- a/lib/friendly_id/migration.rb
+++ b/lib/friendly_id/migration.rb
@@ -1,8 +1,6 @@
-require 'rails'
-
 migration_class =
-  if Rails.version.start_with? "5"
-    ActiveRecord::Migration["#{Rails::VERSION::MAJOR}.#{Rails::VERSION::MINOR}"]
+  if ActiveRecord::VERSION::MAJOR >= 5
+    ActiveRecord::Migration["#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}"]
   else
     ActiveRecord::Migration
   end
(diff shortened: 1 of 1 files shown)


lib/friendly_id/migration.rb
```diff
@@ -1,6 +1,8 @@
+require 'rails'
+
 migration_class =
-  if ActiveRecord::VERSION::MAJOR >= 5
-    ActiveRecord::Migration[4.2]
+  if Rails.version.start_with? "5"
```
parndt: Won’t this break again with Rails 6?

an-nasir: @parndt yea, nice pick. We can make it to check for starts with 5 OR > 5.

But Travis was failing that "Rails is not available" in next line Rails::Version.. 
Can you guide me on requiring rails at this point?

an-nasir: That would be fine too, http://api.rubyonrails.org/classes/ActiveRecord/VERSION.html

an-nasir:
Hi @parndt I've changed Rails to ActiveRecord but I'm getting this error in Travis CI
`Gem::LoadError: Specified 'postgresql' for database adapter, but the gem is not loaded. Add `gem 'pg'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord)`
Any guide, I think I have to change something in Gemfile.rails.x.x.rb but I'm not sure what?

parndt:
@an-nasir can I get your thoughts on this alternative PR #859  please ?

an-nasir:
This is to achieve different thing like it  won't add specific migration in
inheritance. And in rails 5.x it's required to inherit from a specific
migration rather than just Migration.
Lemme know your thoughts

parndt:
Thanks 👍 merged to 9cd4ce2

PR #860, opened 2018-03-09
Fix README.md typo

AlexZaytsav:
I assume that it's a typo and here should be 'rake db:migrate'.

Thanks,
Alex.


parndt:
It used to be rake but this was changed by Rails to use their command instead.  Thanks!

PR #862, opened 2018-04-19
Remove deprecations for Rails 5.1+

yjukaku:
This favors using `will_save_change_to_attribute?` vs the old `attribute_changed?` methods which are now deprecated.

Closes #816


lazyatom:
Apologies for the delay in me looking at this. I think the main thing missing here is a test that demonstrates the need. After a little exploration, I believe this demonstrates both the deprecation issue, and the behaviour change issue:

```ruby
# in test/slugged_test.rb, at the end of SluggedTest
  test "should not raise deprecation warnings when using old slug within callback" do
    klass = Class.new model_class do
      friendly_id :name, :use => :slugged

      attr_accessor :slug_value_in_callback

      after_save do
        self.slug_value_in_callback = to_param
      end

      def self.name
        "Journalist"
      end
    end

    transaction do
      instance = klass.new :name => 'foo'
      assert instance.save
      assert instance.valid?
      instance.name = 'foobar'
      instance.slug = nil
      instance.save
      assert_equal 'foobar', instance.slug
      assert_equal 'foobar', instance.slug_value_in_callback
    end
  end
```

Testing this locally, under Rails 5.2, everything works at the test passes, but under Rails 5.1, we get a bunch of deprecation warnings _and_ the test fails: within the `after_save` callback, instead of using the new `'foobar'` slug, we get the old `'foo'` slug.

I _think_ this test demonstrates that even though the build is passing for Rails 5.2, the behaviour is actually broken under Rails 5.1, and we need to adopt a fix like yours. Would you mind adding this test to your pull request?

yjukaku:
Sure, I'll add that test. Is there an easy way to test the gem on different Rails versions?

lazyatom:
I've been running them by setting the Gemfile via an environment variable, e.g.

```shell
BUNDLE_GEMFILE=gemfiles/Gemfile.rails-5.1.rb bundle install
BUNDLE_GEMFILE=gemfiles/Gemfile.rails-5.2.rb bundle install
BUNDLE_GEMFILE=gemfiles/Gemfile.rails-5.1.rb bundle exec ruby -Itest test/slugged_test.rb
BUNDLE_GEMFILE=gemfiles/Gemfile.rails-5.2.rb bundle exec ruby -Itest test/slugged_test.rb
```

yjukaku:
I added the test (and fixed a small bug with this PR).

I'm not 100% sure I understand what the test is for, but I seem to have it failing on this branch only in Rails 5.0, but not 5.1 and 5.2. 🤷‍♂️

saiqulhaq:
@yjukaku you can use this gem to test in many rails version, it's pretty easy to use
https://github.com/thoughtbot/appraisal

lazyatom:
@yjukaku I've extracted a version of this test, along with a bunch of other regression tests, into #867, and then added my "fix" (delete some code) as a separate commit. Let me know what you think.

yjukaku:
Sounds good. 😆

PR #863, opened 2018-04-20
Fix build and test against Rails 5.1 and 5.2

jhawthorn:
This fixes some small issues with the build and adds gemfiles to test against Rails 5.1 and 5.2.

It would be wonderful if we could get a new gem release to support Rails 5.2 :pray: :gem: 

---

EDIT:
I'm still seeing some errors on Rails 4.0 and 4.1. A git bisect suggests this may be introduced by 4bd4300035b5c250aeb2e5feec4c2feb9bcf2a19. I'm investigating further :mag: . (but maybe it is time to drop 4.0 and 4.1)

EDIT2: This is fixed by reverting 4bd4300035b5c250aeb2e5feec4c2feb9bcf2a19. Not sure why.

---

* [x] Fix mysql dep on Rails 4.2
* [x] Fix pg dep on Rails 4.x
* [x] Add gemfiles for Rails 5.1 and 5.2
* [x] Investigate why defering ActiveRecord patch breaks Rails 4.0 and 4.1
* [x] Fix jruby jdbc issues

### DIFF
--- a/lib/friendly_id.rb
+++ b/lib/friendly_id.rb
@@ -4,7 +4,6 @@ require "friendly_id/base"
 require "friendly_id/object_utils"
 require "friendly_id/configuration"
 require "friendly_id/finder_methods"
-require 'friendly_id/railtie' if defined?(Rails)
 
 =begin
 
--- a/lib/friendly_id/object_utils.rb
+++ b/lib/friendly_id/object_utils.rb
@@ -70,3 +70,7 @@ Object.send :include, FriendlyId::ObjectUtils
 # one of its descendants.
 
 FriendlyId::UNFRIENDLY_CLASSES.each { |klass| FriendlyId.mark_as_unfriendly(klass) }
+
+ActiveSupport.on_load(:active_record) do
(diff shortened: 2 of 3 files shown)


jhawthorn:
I believe this is ready to go! :shipit: 

A quick summary of changes:
* Adjust `gemfiles` to specify older `pg` and `mysql2` gems where needed (ActiveRecord has a non-explicit dependency on these, and older AR versions need older versions of these gems)
* Add some jruby builds to `allowed_failures`. The jdbc adapters just aren't ready for the latest rails versions (even latest patch versions of 5.0 and 5.1 :disappointed:)
* Adjust the change made by 4bd4300035b5c250aeb2e5feec4c2feb9bcf2a19 to be compatible with Rails 4.0 and 4.1 (the build was already failing when that was merged so this wasn't spotted). These versions aren't supported by Rails itself anymore, so `friendly_id` could consider dropping support as well.
* Add tests for Rails 5.1 and 5.2 to TravisCI :muscle: 

@parndt could I trouble you for a review and hopefully a new gem release? Thanks so much :bowing_man:

PR #868, opened 2018-05-17
Remove temp solution for Rails 5 from README.md

lujanfernaud:
Migration warnings on Rails 5 were fixed in https://github.com/norman/friendly_id/commit/2a7121df300e0acc2e18c75f83ad9e17152f5a9f

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,9 +54,6 @@ Generate the friendly configuration file
 rails generate friendly_id
 ```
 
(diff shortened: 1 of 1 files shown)


lujanfernaud:
@zhisme I don't receive any warnings or errors when running the migration on a Rails 5.2.0 app with FriendlyId 5.2.4.

selection_139

Issue #869, opened 2018-05-18
Directly inheriting from ActiveRecord::Migration is not supported


chrisnopa:
Hey everyone,
First, I tried to solve this error on my own, but despite all my efforts, I'm still not sure of what fails exactly when I try to run `db:migrate`! I keep getting this error message:
```rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateFriendlyIdSlugs < ActiveRecord::Migration[4.2]```   
I am using Rails 5.2 version and i checked my gemfile, here is what I find:
```
gem 'friendly_id', '~> 5.1.0', :require => 'friendly_id'
ruby '2.4.3'

Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
gem 'rails', '~> 5.2.0'
```

It looks like it's finding Rails 4.2 somewhere, and It breaks because of this, but I can't find where/why.
Is there any problem with the 5.2 rails version?
Thanks for help,

C.

louishoang:
@chrisnopa After generating the friendly configuration file using`rails generate friendly_id`, there is a migration file that looks similar to this `db/migrate/2018xxxxx_create_friendly_id_slugs.rb`.

At the beginning of the file you will see `CreateFriendlyIdSlugs < ActiveRecord::Migration`, please replace it with `CreateFriendlyIdSlugs < ActiveRecord::Migration[5.2]`. It worked on my end.

jcoyne:
Duplicate of #805

PR #874, opened 2018-06-22
Fixes for 'Directly inheriting from ActiveRecord::Migration is not supported' errors

### LINKED ISSUE #805: Error: Directly inheriting from ActiveRecord::Migration is not supported.
After installing the gem, I encountered the following error when attempting to run `rails db:migrate`:

> rails aborted!
> StandardError: An error has occurred, this and all later migrations canceled:
> 
> Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:
> 
>   class CreateFriendlyIdSlugs < ActiveRecord::Migration[4.2]

The way to fix the error was to put `[5.1]` at the end of `ActiveRecord::Migration`.

k1bs:
resolves #805 and #869 

After encountering this error today, these changes pass the variable of the rails version to the template if Rails version > 5.

Tested with Rails version 5.2, `db:migrate` completes with no errors

### DIFF
--- a/lib/friendly_id/migration.rb
+++ b/lib/friendly_id/migration.rb
@@ -1,11 +1,4 @@
-MIGRATION_CLASS =
-  if ActiveRecord::VERSION::MAJOR >= 5
-    ActiveRecord::Migration["#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}"]
-  else
-    ActiveRecord::Migration
-  end
-
-class CreateFriendlyIdSlugs < MIGRATION_CLASS
+class CreateFriendlyIdSlugs < ActiveRecord::Migration<%= migration_version %>
   def change
     create_table :friendly_id_slugs do |t|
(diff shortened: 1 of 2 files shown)


k1bs:
Interesting.

My solution was to use methods to create an instance variable for the generation of the migration. The `migration_template` method's template accepts ERB, which I used to dynamically generate a version number for the aforementioned 'Directly inheriting from ActiveRecord::Migration is not supported' errors that arose in Rails versions 5.1+.

Despite the fact that my solution solves the issue for both Rails 4 and 5, it's now causing the tests to fail, as I'm not sure they're designed to parse the ERB.

As a temporary solution, I tried renaming the migration template as a `.tt` file (like rails itself does), but this only breaks it further.

To be honest, I'm not really sure how to proceed since I don't fully understand your tests.

parndt:
@k1bs I think you could call the migration file `migration.rb.erb`?

PR #877, opened 2018-07-19
fix History + SequentiallySlugged + Scope issues

kshnurov:
`SequentiallySlugged` with `History` doesn't look into previously used slugs, which can result in a wrong `last_sequence_number` and failure.

### DIFF
--- a/lib/friendly_id/sequentially_slugged.rb
+++ b/lib/friendly_id/sequentially_slugged.rb
@@ -7,17 +7,11 @@ module FriendlyId
     def resolve_friendly_id_conflict(candidate_slugs)
       candidate = candidate_slugs.first
       return if candidate.nil?
-      scope = scope_for_slug_generator
-      if friendly_id_config.uses?(:history)
-        base_class = Slug
-      else
-        base_class = self.class.base_class
-      end
-      SequentialSlugCalculator.new(scope,
+      SequentialSlugCalculator.new(scope_for_slug_generator,
                                   candidate,
                                   friendly_id_config.slug_column,
                                   friendly_id_config.sequence_separator,
-                                  base_class).next_slug
+                                  slug_base_class).next_slug
     end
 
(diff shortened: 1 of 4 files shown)


kshnurov:
@norman @parndt could you please look into this?

lib/friendly_id/sequentially_slugged.rb
```diff
@@ -7,11 +7,17 @@ def self.setup(model_class)
     def resolve_friendly_id_conflict(candidate_slugs)
       candidate = candidate_slugs.first
       return if candidate.nil?
-      SequentialSlugCalculator.new(scope_for_slug_generator,
+      scope = scope_for_slug_generator
+      if friendly_id_config.uses?(:history)
+        base_class = Slug
+      else
+        base_class = self.class.base_class
+      end
```
parndt: could you extract this into a method `base_class`?

lib/friendly_id/sequentially_slugged.rb
```diff
@@ -7,11 +7,17 @@ def self.setup(model_class)
     def resolve_friendly_id_conflict(candidate_slugs)
       candidate = candidate_slugs.first
       return if candidate.nil?
-      SequentialSlugCalculator.new(scope_for_slug_generator,
+      scope = scope_for_slug_generator
```
parndt: this could probably go back to being used inline?

```
SequentialSlugCalculator.new(scope_for_slug_generator,

lib/friendly_id/sequentially_slugged.rb
kshnurov: @parndt done

lib/friendly_id/sequentially_slugged.rb
kshnurov: @parndt done

parndt:
Thanks a lot @kshnurov 👍

Issue #880, opened 2018-08-28
ActiveRecord::RecordNotFound attributes are not passed


sPaCeMoNk3yIam:
https://github.com/norman/friendly_id/blob/e8186c6d407ba63ac6edfeff251c7355f8e212dc/lib/friendly_id/finder_methods.rb#L23
https://github.com/norman/friendly_id/blob/e8186c6d407ba63ac6edfeff251c7355f8e212dc/lib/friendly_id/finder_methods.rb#L37
https://api.rubyonrails.org/classes/ActiveRecord/RecordNotFound.html

FriendlyId should pass all attributes (`model`, `primary_key`, `id`) correctly to ensure consistent behavior for error handling. With `id`, of course, being the `friendly_id`.

parndt:
Fixed by #890

PR #881, opened 2018-08-28
Add more context to RecordNotFound exception

dpuscher:
[`ActiveRecord::RecordNotFound`](https://api.rubyonrails.org/classes/ActiveRecord/RecordNotFound.html) has the ability to add more context to the error using three additional parameters: `model`, `primary_key` and `id`.

Using them makes it more easy to analyze the exception and make the application behave in a smarter way (e.g. redirect to another page when objects are deleted from database)

This should fix bug #880.

### DIFF
--- a/lib/friendly_id/finder_methods.rb
+++ b/lib/friendly_id/finder_methods.rb
@@ -20,7 +20,7 @@ module FriendlyId
       return super if args.count != 1 || id.unfriendly_id?
       first_by_friendly_id(id).tap {|result| return result unless result.nil?}
       return super if potential_primary_key?(id)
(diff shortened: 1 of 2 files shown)


sPaCeMoNk3yIam:
Seems like, despite https://github.com/rails/rails/commit/d763956ed95bce581f12151a5ed4df12bcefdeee was written 3 years ago, it was only released with Rails v5.0.0.beta1. So that fix would mean not being compatible to Rails prior to v5.x.

parndt:
Thank you for the PR, I've since merged another PR which also addresses this issue #890

PR #898, opened 2019-03-15
Temporarily direct bugreporting info link to something more appropriate

SkyWriter:
Seems like the original site has been hacked or something.

I suggest we change the link until it's back.


parndt:
Thanks - how unfortunate

PR #903, opened 2019-04-28
Fix/Slug Validation in unset_slug_if_invalid

### LINKED ISSUE #902: slug being unset if model has any errors (unset_slug_if_invalid)
When running a .save() method on a model which has errors on other fields (e.g.` name` is invalid), the `slug` gets set to nil but was valid. It should preserve the value after .save() on the model so the user can fix the remaining errors without having to rewrite the slug. This also leads to other errors on the slug field that may have caused the model to fail validation to be removed.

method `unset_slug_if_invalid` shouldn't be checking for `errors.present?`. Instead it should only check that the `slug` field has any errors (doing `errors[friendly_id_config.query_field].present?` solves the problem)

```ruby
  def unset_slug_if_invalid
    if errors[friendly_id_config.query_field].present? && attribute_changed?(friendly_id_config.query_field.to_s)
      diff = changes[friendly_id_config.query_field]
      send "#{friendly_id_config.slug_column}=", diff.first
    end
   end
```

Ancez:
updated `unset_slug_if_invalid` to only validate errors for the slug field instead of all model fields.

closes #902

### DIFF
--- a/lib/friendly_id/slugged.rb
+++ b/lib/friendly_id/slugged.rb
@@ -375,7 +375,7 @@ Github issue](https://github.com/norman/friendly_id/issues/185) for discussion.
     private :slug_generator
 
     def unset_slug_if_invalid
-      if errors.present? && attribute_changed?(friendly_id_config.query_field.to_s)
+      if errors[friendly_id_config.query_field].present? && attribute_changed?(friendly_id_config.query_field.to_s)
         diff = changes[friendly_id_config.query_field]
         send "#{friendly_id_config.slug_column}=", diff.first
       end


parndt:
@Ancez hi! Could you please pull in the latest changes from master into this PR?

Ancez:
> @Ancez hi! Could you please pull in the latest changes from master into this PR?

Will do later on today, sorry for the late response

nedenwalker:
Hi there, we encountered the same issue today that this PR purports to fix. We were wondering about the state of it. Is there something we can do to help?

parndt:
@nedenwalker It needs rebasing against master because it's failing CI on its current HEAD.

Does the change fix your problem?

nedenwalker:
@parndt I can confirm that the fix does eliminate the bug. When the form reloads with previous form values and error messages, the slug field is now also present.

nedenwalker:
Wasn't sure on how to run teamcity? tests locally or rebase someone else's fork etc, so made my own fork, made the code change and submitted a new PR to get CI running in your system: https://github.com/norman/friendly_id/pull/936

nedenwalker:
Figured out local testing issue, was running `bundle exec rails test`, needed to instead do `bundle exec rake test`. I see lots of failures. Will need some time to process it all, since change was a one-liner, hopefully fix to tests is also a one liner.

nedenwalker:
@parndt I see 5 test failures, all related to the code change. I cannot admit to understanding the coding enough, but in particular, I see one test that specifies the code should do what it used to do, ie wipe the slug when there is another field validation failure. This code change would alter that contract. https://github.com/norman/friendly_id/blob/master/test/slugged_test.rb#L95 Will leave it to you to determine if the functionality should change or not.

Ancez:
yeah there are some tests to fix otherwise this fixes the issue of the slug being removed. If you want to you can fork this line and fix the specs for it. If not then I'll fix these specs tonight if that's alright, until then I recommend changing the base within your projects Gemfile for this gem to this PR

khng:
Fixed the test and submitted a new PR here: https://github.com/norman/friendly_id/pull/938.

PR #904, opened 2019-05-09
Fixes db:migrate command in Readme

zykadelic:
Should be `rake`, not `rails` 🙂


parndt:
Thank you, but this is correct according to Rails docs: https://guides.rubyonrails.org/active_record_migrations.html#running-migrations

PR #907, opened 2019-06-21
Ability to change slugs manually take 2

kwent:
Cf. https://github.com/norman/friendly_id/pull/758

### DIFF
--- a/lib/friendly_id/slugged.rb
+++ b/lib/friendly_id/slugged.rb
@@ -293,6 +293,15 @@ Github issue](https://github.com/norman/friendly_id/issues/185) for discussion.
       value
     end
 
+    # Whether to allow changing slug manually
+    #
+    # This method can be overrided in your model. By default you can not
+    # change slug through `#update_attributes`. But you can set on that by, for example,
+    # simple check `self.slug_changed?`
+    def should_manually_change_friendly_id?
+      false
+    end
+
     # Whether to generate a new slug.
     #
     # You can override this method in your model if, for example, you only want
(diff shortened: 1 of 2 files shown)


lib/friendly_id/slugged.rb
```diff
@@ -357,6 +366,10 @@ def set_slug(normalized_slug = nil)
         candidates = FriendlyId::Candidates.new(self, normalized_slug || send(friendly_id_config.base))
         slug = slug_generator.generate(candidates) || resolve_friendly_id_conflict(candidates)
         send "#{friendly_id_config.slug_column}=", slug
+      elsif should_manually_change_friendly_id?
+        candidates = FriendlyId::Candidates.new(self, self.slug)
+        slug = slug_generator.generate(candidates) || resolve_friendly_id_conflict(candidates)
+        send "#{friendly_id_config.slug_column}=", slug
```
parndt: this could probably be extracted as common functionality:

```ruby
    # Sets the slug.
    def set_slug(normalized_slug = nil)
      if should_generate_new_friendly_id?
        set_slug_with_candidate(normalized_slug || send(friendly_id_config.base))
      elsif should_manually_change_friendly_id?
        set_slug_with_candidate(self.slug)
      end
    end
    private :set_slug

    def set_slug_with_candidate(candidate)
      candidates = FriendlyId::Candidates.new(self, candidate)
      slug = slug_generator.generate(candidates) || resolve_friendly_id_conflict(candidates)
      send "#{friendly_id_config.slug_column}=", slug
    end
    private :set_slug_with_candidate
```

What do you think?

PR #916, opened 2019-08-26
Record history when scope changes but slug does not

glampr:
In order to record history, we should check if there is a new slug.
If the record uses the scoped module we should also check if there is **a change in the scope**.

### DIFF
--- a/lib/friendly_id/history.rb
+++ b/lib/friendly_id/history.rb
@@ -141,6 +141,7 @@ method.
       if friendly_id_config.uses?(:scoped)
         check = check && latest_history.scope == serialized_scope
       end
+      check
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/friendly_id/history.rb
```diff
@@ -134,5 +134,13 @@ def create_slug
...
       end
     end
+
+    def history_is_up_to_date?
+      latest_history = slugs.first
+      check = latest_history.try(:slug) == friendly_id
+      if friendly_id_config.uses?(:scoped)
+        check = check && latest_history.scope == serialized_scope
+      end
+    end
```
parndt: you forgot to return `check` on the last line of this method, so if the config does not use scoped then the method will return `nil`

```suggestion
      check
    end
```

glampr:
@parndt thanks for the review. I have implemented your requested change.

parndt:
Thanks @glampr, it makes sense and I've merged it.

PR #924, opened 2019-11-19
use destroy_all rather than delete_all when creating historical…

alexdean:
Our application adds an `after_commit` callback to `FriendlyId::Slug`. The usage of `delete_all` rather than `destroy_all` in `FriendlyId::History::FinderMethods#create_slug` means that this callback is not always invoked.

This PR changes this `delete_all` to a `destroy_all`, which will invoke callbacks for the deleted records. Performance impact should be minimal given that (at most) 1 record is loaded and then deleted.

Additionally: I noticed that no tests failed when i removed the `delete_all` call entirely. I think the test I've added will guard against that.

### DIFF
--- a/lib/friendly_id/history.rb
+++ b/lib/friendly_id/history.rb
@@ -1,4 +1,3 @@
-
 module FriendlyId
 
 =begin
--- a/lib/friendly_id/history.rb
+++ b/lib/friendly_id/history.rb
@@ -1,3 +1,4 @@
+
 module FriendlyId
 
 =begin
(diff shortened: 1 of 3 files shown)


lib/friendly_id/history.rb
```diff
@@ -1,3 +1,4 @@
+
```
parndt: Typo?
```suggestion
```

alexdean: huh. yeah, i agree that's a fat-finger error. will fix.

parndt:
Thank you @alexdean ! 👍 🥇

alexdean:
happy to help!

PR #925, opened 2019-11-24
Updating slugged normalize method to preserve case

swaathi:
This PR provides the ability for the `normalize_friendly_id` method to preserve case when parameterizing slug value. This allows for cleaner URLs where sometimes letter case carries meaning.


parndt:
thanks very much for this, however this is a breaking change and, as you can override the method in your own model to achieve this, I'm going to close the pull request.

PR #931, opened 2019-12-26
fix: Scoped#should_generate_new_friendly_id? only for persisted record

beauraF:
Since the introduction of `Scoped#should_generate_new_friendly_id?`, it is no longer possible to create a record with a predefined slug.

I suggest to check if it is a `new_record`. But maybe it's more relevant to see if `slug_changed?` ?


beauraF:
I'm working on adding a test in the coming days.

PR #940, opened 2020-04-11
Fix mysql jobs in CI build matrix

mattbrictson:
The mysql jobs are failing in CI. It seems this is because there is no mysql service running. Fix by adding a mysql service to the CI config.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
         ruby-version: ${{ matrix.ruby }}
         version: ${{ matrix.ruby }}
     - uses: actions/checkout@v1
-    - run: sudo apt-get update && sudo apt-get install libpq-dev postgresql-client libmysqlclient-dev libsqlite3-dev -y
+    - run: sudo apt-get update && sudo apt-get install libpq-dev postgresql-client libmysqlclient-dev mysql-client libsqlite3-dev -y
     - id: cache-bundler
       uses: actions/cache@v1
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         ruby-version: ${{ matrix.ruby }}
         version: ${{ matrix.ruby }}
     - uses: actions/checkout@v1
-    - run: sudo apt-get update && sudo apt-get install libpq-dev postgresql-client libmysqlclient-dev libsqlite3-dev -y
+    - run: sudo apt-get update && sudo apt-get install libpq-dev postgresql-client libmysqlclient-dev mysql-client libsqlite3-dev -y
```
mattbrictson, line 26: `mysql-client` is needed because `rake db:up` uses the `mysql` CLI.

Rakefile
```diff
@@ -71,7 +71,7 @@ namespace :db do
     driver = FriendlyId::Test::Database.driver
     config = FriendlyId::Test::Database.config[driver]
     commands = {
-      "mysql"    => "mysql -u #{config['username']} --password=#{config['password']} -e 'create database #{config["database"]};' >/dev/null",
+      "mysql"    => "mysql -h #{config['host']} -P #{config['port']} -u #{config['username']} --password=#{config['password']} -e 'create database #{config["database"]};' >/dev/null",
```
mattbrictson, line 74: I had to add host and port, otherwise `mysql` would try (and fail) to connect to `/var/lib/mysql/mysql.sock`.

parndt:
Amazing thank you so much!!

Issue #947, opened 2020-08-02
Saving record with `save(validate: false)` skips generating slug, sometimes


vfonic:
I just discovered an issue that happened in my app.

When I call `record.save!(validate: false)` on a record, it doesn't generate slug for it.

However, if I call `record.valid?` before calling `record.save!(validate: false)`, it does generate the slug.

This is what made me miss this issue at first.

I believe this could be a bug within the friendly_id gem.

parndt:
Thanks @vfonic can you please post a minimal reproduction case, or test case?

vfonic:
@parndt here's the minimal repro case:

https://github.com/vfonic/friendly_id/commit/22d923b6743f62909c225efb81cb14788e04ee37

It's on this branch on my fork (just 1 commit):
https://github.com/vfonic/friendly_id/tree/generating-slug-without-validation

Here's the output of running the tests:

```
```

vfonic:
@parndt did you have any chance to look into this one? I don't want this issue to get closed because it's "stale".

Thanks! 🙏

parndt:
> @parndt did you have any chance to look into this one? I don't want this issue to get closed because it's "stale".

@vfonic sorry, not yet, available time is so low right now ☹️ 

It's nothing to do with what #952 fixes is it?

vfonic:
@parndt I rebased on top of the `master` the test I added in my fork and it's still failing.

Going through the code, this line seems to be the issue:
https://github.com/norman/friendly_id/blob/master/lib/friendly_id/slugged.rb#L250

```rb
model_class.before_validation :set_slug
```

The slug is only set if validation occurs. So if you do `record.save(validate: false)`,  the validation will never happen, therefore slug will never be set.

vfonic:
@parndt I've force-pushed my fix to `vfonic:generating-slug-without-validation` and realized there's already a PR open:
https://github.com/norman/friendly_id/pull/948

The build for this PR is now passing.

parndt:
Sounds good to me @vfonic. I've merged it so this is closed by #948

vfonic:
Thank you, Philip! 👍

PR #948, opened 2020-08-12
Generating slug when validation is skipped

parndt:
This opens the change up for discussion and continuous integration.

Currently, this change brings in a failing test by @vfonic so that we can
get it passing.


vfonic:
Related issue: https://github.com/norman/friendly_id/issues/947

parndt:
Thanks @vfonic !!

PR #951, opened 2020-08-16
Add `parse_friendly_id` to make slug finder downcase opt-in

### LINKED ISSUE #950: V 5.4.0 breaking change to Object.friendly.find method
Before merge of 787, one could pass a mixed-case slug to friendly.find as such:

`Customer.friendly.find(params[:id`])

We don't use the customer name for the slug; rather we used SecureRandom.urlsafe_base64 to generate a random string.  Thus the slugs for our customers contain a mix of upper and lowercase characters.  At this point we have thousands of customers with mixed-case slugs in our databases.

Because this is breaking (for us) some kind of opt-in to the new behavior would seem to make the most sense.

parndt:
In #787 we merged a change which found slugs by `id.downcase` by
default, but this has understandably caused issues for other users of
FriendlyId who don't want to do this.

This reverts that change and introduces a new method, 
`parse_friendly_id`, which can be overridden similar to 
`normalize_friendly_id`.  

Instead of being used for slug generation, `parse_friendly_id` is used
for parsing incoming slugs.

Fixes #950

### DIFF
--- a/lib/friendly_id/finder_methods.rb
+++ b/lib/friendly_id/finder_methods.rb
@@ -59,7 +59,7 @@ module FriendlyId
     end
 
     def first_by_friendly_id(id)
-      find_by(friendly_id_config.query_field => id.downcase)
+      find_by(friendly_id_config.query_field => id)
     end
 
     def raise_not_found_exception(id)
--- a/lib/friendly_id/finder_methods.rb
+++ b/lib/friendly_id/finder_methods.rb
@@ -20,8 +20,8 @@ module FriendlyId
       return super if args.count != 1 || id.unfriendly_id?
       first_by_friendly_id(id).tap {|result| return result unless result.nil?}
       return super if potential_primary_key?(id)
-      raise_not_found_exception id
-      
+
+      raise_not_found_exception(id)
     end
 
     # Returns true if a record with the given id exists.
@@ -39,7 +39,7 @@ module FriendlyId
     end
 
     def exists_by_friendly_id?(id)
-      where(friendly_id_config.query_field => id).exists?
+      where(friendly_id_config.query_field => parse_friendly_id(id)).exists?
     end
 
     private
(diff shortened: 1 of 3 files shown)


parndt:
@ghbooth12 @vfonic @ccasabona @coneybeare @marckohlbrugge

parndt:
I haven't laboured over testing for this yet, I wanted to know if this would solve the problem for everyone?

lib/friendly_id/finder_methods.rb
```diff
@@ -59,14 +59,47 @@ def potential_primary_key?(id)
...
 
     def first_by_friendly_id(id)
-      find_by(friendly_id_config.query_field => id.downcase)
+      find_by(friendly_id_config.query_field => parse_friendly_id(id))
+    end
+
+    # Parse the given value to make it suitable for use as a slug according to
+    # your application's rules.
+    #
+    # This method is not intended to be invoked directly; FriendlyId uses it
```
vfonic, line 68: I'd say that this paragraph is not needed. I'd leave it out. I don't think there's much value to have this extra paragraph here and maybe it's also incorrect, as maybe someone might want to use `parse_friendly_id` in some other situation...

vfonic: If this paragraph is removed, I'd also remove `However, ` from the next paragraph

lib/friendly_id/finder_methods.rb
```diff
@@ -59,14 +59,47 @@ def potential_primary_key?(id)
...
+    #       friendly_id :name_and_location
+    #
+    #       def name_and_location
+    #         "#{name} from #{location}"
+    #       end
+    #
+    #       # Use default slug, but lower case
+    #       # If `id` is "Jane-Doe" or "JANE-DOE", this finds data by "jane-doe"
+    #       def parse_friendly_id(slug)
+    #         super.downcase
```
vfonic, line 88: This is an interesting usage. It could be better to use the passed-in (original) `slug` argument. But maybe someone would like to compose the parsing with some other modules they wrote. 🤷

lib/friendly_id/finder_methods.rb
coneybeare: The default implementation is just a passthrough accessor to `value`, so I would argue that this paragraph is warranted. It also has [precedence](https://github.com/norman/friendly_id/blob/f5733d024da4760fba64ce9ce6d3ad5fc0ee5e25/lib/friendly_id/slugged.rb#L256)

lib/friendly_id/finder_methods.rb
coneybeare: This method is used by the finder, meaning you don't have the original at this point to query, only the slug from the url

lib/friendly_id/finder_methods.rb
vfonic: Ah interesting. Thanks!

lib/friendly_id/finder_methods.rb
vfonic: Yeah, that's what I meant by original: originally received from the user

ccasabona:
Thank you for looking in to this.

The fixes in PR 951 fix the regression to Object.friendly.find and again permit existing slugs to be mixed-case.  I can detect no other changes across our projects.

Charlie Casabona

rsslldnphy:
hi @parndt , what kind of help are you after?

parndt:
> hi @parndt , what kind of help are you after?

thank you for asking @rsslldnphy - basically I haven't allocated time to add specs for this change yet; if anyone else can add specs that would help to avoid regressions then we can merge this really soon and release a new version.

parndt:
In particular I deleted a bunch of specs but should they still run if the opt-in is activated ?

Also keen for opinions on whether the current specs cover our code enough considering that anyone can override `parse_friendly_id` for any purpose.

rsslldnphy:
I'd argue that the same is true for _any_ method - anyone can override them for any purpose if they want. And if they do I think it's up to them to test that the method they've written does what they expect it to.

I can see an argument for having a spec for this specific case if it's an example that will be used in the docs. Effectively that would be saying that "overriding this specific method is part of the libraries' public API".

However I also think there's an argument that as this PR effectively removes functionality and instead suggests that users add it themselves, it makes sense that the specs are removed along with it. There's no way for you to use this functionality without writing your own code, and if you're writing your own code you should be testing it.

That being said I can probably find some time in the next few days to add the specs back in if needed.

xtagon:
👍 Would love to see this released now that it is merged. Updating to 5.4.0 is necessary for Ruby 2.7 support but without this revert it also introduces a breaking change.

parndt:
Ah, I'd thought this made it into `5.4.1`, but I see that it missed it by a couple of days.. I'll get 5.4.2 released shortly.

xtagon:
Thank you so much!

parndt:
> Thank you so much!

you're welcome; it's released now, so you can use it from rubygems.org 😄

PR #952, opened 2020-08-27
Remove unexpected :slug error on valid model, fixes #949

### LINKED ISSUE #949: Unexpected :slug error on valid, unpersisted model
While testing friendly_id 5.4.0, we are seeing something that is odd (I think -- I may be misunderstanding).

After constructing (but not saving) an instance (let's call it `my_instance`) of a given MyModel that uses `friendly_id :some_method, use: [:slugged, :finders]`, we see that while `my_instance.valid?` is `true`, `my_instance.errors.messages` is `{:slug=>[]}`.

Is this intentional? I find it rather peculiar, but I cannot pretend to be an expert here or fully understand what's going on under the hood with friendly_id. I am happy to flesh out this example if desired -- just let me know what extra info, if any, would be helpful!

caiofilipemr:
I believe this simple change solves the _issue_ #949.

### DIFF
--- a/lib/friendly_id/slugged.rb
+++ b/lib/friendly_id/slugged.rb
@@ -375,7 +375,7 @@ Github issue](https://github.com/norman/friendly_id/issues/185) for discussion.
     private :slug_generator
 
     def unset_slug_if_invalid
-      if errors[friendly_id_config.query_field].present? && attribute_changed?(friendly_id_config.query_field.to_s)
+      if errors.key?(friendly_id_config.query_field) && attribute_changed?(friendly_id_config.query_field.to_s)
         diff = changes[friendly_id_config.query_field]
         send "#{friendly_id_config.slug_column}=", diff.first
       end


parndt:
Thanks @caiofilipemr that looks like it could do the trick, but CI is failing for this change?

giovannibonetti:
Hi, @parndt! I work with @caiofilipemr and we were looking into this together.

> but CI is failing for this change?

After [taking a closer look](https://github.com/norman/friendly_id/pull/952/checks?check_run_id=1036629354) it seems like the few cases that failed are not related to these changes at all:

image

Can you run the tests again? It seems like the Coveralls configuration is a bit unstable.

parndt:
> After [taking a closer look](https://github.com/norman/friendly_id/pull/952/checks?check_run_id=1036629354) it seems like the few cases that failed are not related to these changes at all:

it also reports this failure: https://github.com/norman/friendly_id/pull/952/checks?check_run_id=1036629235#step:10:99

I'll re-run and see. It's failing on Rails 5.0 tests which leads me to believe it's related to Rails 5.

parndt:
> It seems like the Coveralls configuration is a bit unstable.

Sometimes I wonder if it's even a worthwhile measurement. 😄

parndt:
Rather than fight CI, I've opened #954

parndt:
@caiofilipemr can you pull in the latest changes to your branch please so that CI may pass?

parndt:
Thanks!

PR #953, opened 2020-08-28
Add FUNDING.yml for sponsorship

parndt:
To enable more time to be spent on this project, it makes sense to
link to existing GitHub sponsors.


parndt:
@norman what do you think of this? :smile:

norman:
I'm 100% fine with it, as long as none of the money goes to me. You're doing to work, you get the funds.

PR #962, opened 2020-12-16
rails-5-2-regression

sbeckeriv:
Dearest Reviewer,
    
    This fork and pr works for our workflow. I am unsure this is the correct
    solution for everyone.
    
    1) revert the changes api changes pr
    2) if there are changes still.. and there is a diff use those first
    3) i believe the slugged_test was updated incorrectly
    
    Becker


sbeckeriv:
i did not mean to make this pr on this repo. very sorry

PR #965, opened 2021-01-24
SimpleI18n: Handle regional locales

xymbol:
It defines how to use slug columns for locales with region subtags.

This could result in a breaking change for people using column names like `title_pt-BR`. Here I try to be explicit about and define how to use subtags, which we currently don't.


parndt:
It makes sense to do this. If we release this as 6.0.0 then it would be OK even if it's a breaking change.

parndt (approved):

xymbol:
@parndt Thanks so much! 🙏

parndt:
likewise @xymbol

PR #972, opened 2021-06-25
Don't call upstream if nothing found via friendly ID search

corwinstephen:
Prevent false positives when friendly ID doesn't match

### DIFF
--- a/lib/friendly_id/finder_methods.rb
+++ b/lib/friendly_id/finder_methods.rb
@@ -28,7 +28,7 @@ module FriendlyId
     def exists?(conditions = :none)
       return super if conditions.unfriendly_id?
       return true if exists_by_friendly_id?(conditions)
-      super
+      false
     end
 
     # Finds exclusively by the friendly id, completely bypassing original


lib/friendly_id/finder_methods.rb
```diff
@@ -28,7 +28,7 @@ def find(*args)
     def exists?(conditions = :none)
       return super if conditions.unfriendly_id?
       return true if exists_by_friendly_id?(conditions)
```
parndt, line 30:
```suggestion
      exists_by_friendly_id?(conditions)
```

lib/friendly_id/finder_methods.rb
```diff
@@ -28,7 +28,7 @@ def find(*args)
...
-      super
+      false
```
parndt, line 31:
```suggestion
```

parndt:
@corwinstephen thanks I added some suggestions; we can convert this from a guard clause to just use the result of `exists_by_friendly_id?` as that is what we're most interested in here.

PR #985, opened 2021-12-21
Drop support for Ruby 2.5

alejandroperea:
Ruby 2.5 arrived to EOL. I think it is the moment to drop the support.

### DIFF
--- a/friendly_id.gemspec
+++ b/friendly_id.gemspec
@@ -12,7 +12,7 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
   s.license = "MIT"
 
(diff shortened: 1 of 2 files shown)


friendly_id.gemspec
```diff
@@ -12,7 +12,7 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
   s.license = "MIT"
 
-  s.required_ruby_version = ">= 2.1.0"
+  s.required_ruby_version = ">= 2.6.0"
```
parndt, line 15:
```suggestion
  s.required_ruby_version = ">= 2.1.0"
```

parndt (approved):

PR #986, opened 2022-01-18
Add Ruby 3.1 to CI

petergoldstein:
This PR adds Ruby 3.1 to CI.

It also cleans up an unnecessary exclusion for Ruby 3.0 / Rails 5.2 (Rails 5.2 was already removed from the CI matrix).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,8 +13,6 @@ jobs:
         gemfile: [ '7.0', '6.1', '6.0' ]
         ruby: [ '2.6', '2.7', '3.0', '3.1' ]
         exclude:
-          - ruby: '3.1'
-            gemfile: '6.0'
           - ruby: '2.6'
             gemfile: '7.0'
       fail-fast: false
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,10 +11,10 @@ jobs:
       matrix:
         database: [ mysql, postgresql ]
         gemfile: [ '7.0', '6.1', '6.0' ]
-        ruby: [ '2.6', '2.7', '3.0' ]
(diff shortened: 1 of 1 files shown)


petergoldstein:
I haven't committed to `friendly_id` in quite some time so I'll ask @parndt to review.

parndt (approved):

.github/workflows/test.yml
```diff
@@ -11,10 +11,10 @@ jobs:
       matrix:
         database: [ mysql, postgresql ]
         gemfile: [ '7.0', '6.1', '6.0' ]
-        ruby: [ '2.6', '2.7', '3.0' ]
+        ruby: [ '2.6', '2.7', '3.0', '3.1' ]
         exclude:
-          - ruby: '3.0'
-            gemfile: '5.2'
+          - ruby: '3.1'
+            gemfile: '6.0'
```
parndt, line 17: Wow! Does Rails 6 really not support Ruby 3.1?

.github/workflows/test.yml
```diff
@@ -11,10 +11,10 @@ jobs:
...
           - ruby: '2.6'
```
parndt, line 18:
```suggestion
          - ruby: '2.6'
```

.github/workflows/test.yml
parndt: it does appear to work 😄

parndt:
Thanks @petergoldstein 🎉

petergoldstein:
@parndt So it's a question of formal support.  Rails 6.0 dropped out of primary support before Ruby 3.1 was released.  But either way for these specs is fine - it seems to run ok in most situations (there are some challenges with argument passing in ActionMailer).

PR #988, opened 2022-03-22
Modify gemspec heredoc to be compatible with earlier Ruby versions

### LINKED ISSUE #987: Heredoc with tilde `<<~` in the `.gemspec` is incompatible with older ruby versions
Hi, my team has the displeasure of working on an older ruby version (`2.2.3`) which does not understand all these fancy new tilde heredocs `<<~`, and that leads to this gem failing to install because the `.gemspec` contains the dreaded syntax ; since the `.gemspec` claims tolerating ruby as old as `2.1.0`, maybe it'd be better to use the older dash heredoc `<<-` or to bump up the requirements to ruby `2.3.0` where that syntax was born?

What do you think?

EDIT: After inspecting the code it appears the only use of the `<<~` is in the `.gemspec`, so I would suggest using the pre-existing `<<-` syntax rather than bump up the ruby version requirement.

Bastes:
Fixes #987

### DIFF
--- a/friendly_id.gemspec
+++ b/friendly_id.gemspec
@@ -25,11 +25,9 @@ Gem::Specification.new do |s|
   s.add_development_dependency "ffaker"
   s.add_development_dependency "simplecov"
 
-  s.description = <<~EOM
-    FriendlyId is the "Swiss Army bulldozer" of slugging and permalink plugins for
-    Active Record. It lets you create pretty URLs and work with human-friendly
-    strings as if they were numeric ids.
-  EOM
+  s.description = "FriendlyId is the \"Swiss Army bulldozer\" of slugging " \
+    "and permalink plugins for Active Record. It lets you create pretty URLs " \
+    "and work with human-friendly strings as if they were numeric ids."
 
   s.cert_chain = [File.expand_path("certs/parndt.pem", __dir__)]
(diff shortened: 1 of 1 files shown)


friendly_id.gemspec
```diff
@@ -25,10 +25,10 @@ Gem::Specification.new do |s|
...
 
-  s.description = <<~EOM
-    FriendlyId is the "Swiss Army bulldozer" of slugging and permalink plugins for
-    Active Record. It lets you create pretty URLs and work with human-friendly
-    strings as if they were numeric ids.
+  s.description = <<-EOM
+FriendlyId is the "Swiss Army bulldozer" of slugging and permalink plugins for
+Active Record. It lets you create pretty URLs and work with human-friendly
+strings as if they were numeric ids.
   EOM
```
parndt, line 32:
```suggestion
  s.description = "FriendlyId is the \"Swiss Army bulldozer\" of slugging and" \
    "permalink plugins for Active Record. It lets you create pretty URLs and " \
    "work with human-friendly strings as if they were numeric ids."
```

What about this version, so that standard doesn't change it back, @Bastes ?

Bastes: Agreed :)

Bastes: And done.

Bastes:
Yay! Thank you :)

PR #991, opened 2022-03-30
Fix collision errors when introducing history plugin after the fact

tarzan:
I've redone the work of #850 

The problem that still exists in this current version is:
1. implement the friendly_id slugs
2. figuring out somewhere along the line that you want the history plugin as well
3. create a new sluggable record that would translate to an already existing slug
4. fails with a database constraint on unique slugs

The method 'scope_for_slug_generator' did a bit too much work with an inner join that made existing records with no entry in the history table being filtered out.

### DIFF
--- a/lib/friendly_id/history.rb
+++ b/lib/friendly_id/history.rb
@@ -106,7 +106,7 @@ module FriendlyId
     # to be conflicts. This will allow a record to revert to a previously
     # used slug.
     def scope_for_slug_generator
-      relation = super.joins(:slugs)
+      relation = super.includes(:slugs)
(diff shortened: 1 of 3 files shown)


parndt:
Thank you for redoing this @tarzan 😄

qagusha:
дд

tarzan:
> Thank you for redoing this @tarzan 😄 

Before this becoming stale again: anything keeping you from merging this?

parndt:
I think it's good @tarzan 

@norman can you think of any gotchas here? I've not used the history plugin before.

Issue #937, opened 2020-03-19
Can we have a find_by, returning nil when not found


fidalgo:
Instead of having a `find` that will throw an exception, it would help to have a `find_by` method, that would have the same behaviour as `find_by` in rails, where it returns the first match or nil?

parndt:
@fidalgo does `find_by(slug: params[:id])` work?

fidalgo:
@parndt 
`find_by(slug: id)`  always return `nil`, when the record exists or not.

I'm using `Model.friendly.find_by`

parndt:
Does the below work, without the friendly filter applied?

```ruby
Model.find_by(slug: params[:id])
```

My reasoning is that `slug` should just be a column on your database table so bypassing this library should give the behaviour that you're after.  I tried this with an app that I have running friendly_id and it works as expected.

fidalgo:
I'm sorry, @parndt but I guess we're having a communication issue here.

With friendly_id gem, I can use the method `find` with either `slug` or `id`, but then the record does not exist, it will raise a record not found an exception.

So my request is to add another method that will mimic the `find_by` in Rails https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by

so instead of raising an exception is will return `nil`.

When I run `Model.find_by(slug: params[:id])` it works as expected, otherwise, it would be a bug on Rails, what I would like would be to be able to use `Model.friendly.find_by(slug: params[:id])` that would return the record or `nil`.

Please let me know if I'm missing something!

parndt:
@fidalgo understood, I think. I'm mostly wondering why we need to add that when Rails does it already? 😄 it seems like FriendlyId doesn't need a method for doing what Rails does already. That is why we put our functionality behind `.friendly` so that you still have Rails' methods available.

joemsak:
I found this issue (sup, @parndt how you been) looking for something similar:

For me the idea would be that you don't know if you have primary key ID or slug friendly ID and still want a way to use them interchangeably and just get a nil instead of a raised error:

```ruby
MyModel.friendly.find_safe(id_or_slug) 
# not saying the method has to be called `find_safe()` but hopefully you get the idea
```

Does that make sense?

parndt:
@joemsak I think the idea does make sense in principle, and we could probably support it using something like `MyModel.friendly.find(id_or_slug, raise: false)`

Super nice to hear from you too, Joe! ❤️

joemsak:
[...]

PR #995, opened 2022-07-04
Add `allow_nil: true` option to the finder

joemsak:
This PR adds an optional `allow_nil: true` option to the finder

From discussion in #937 the idea is to add functionality that behaves *like* Rails' `find_by` which returns `nil` when a record is not found. 

This is useful in conditions where the developer is allowing the primary key ID and the friendly slug ID but is not sure if the record will actually be found, and wants a nil instead of a raised exception.

### Usage

```ruby
MyModel.friendly.find("friendly-slug")
# => still works as expected, raises an ActiveRecord::RecordNotFound exception

MyModel.friendly.find("friendly-slug", allow_nil: true)
# => same as above, but some devs may find using this option to be more self-documenting of their intentions
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,12 +116,12 @@ avoid raising `ActiveRecord::RecordNotFound` and accept a `nil`.
 ```ruby
 MyModel.friendly.find("bad-slug") # where bad-slug is not a valid slug
 MyModel.friendly.find(123)        # where 123 is not a valid primary key ID
-MyModel.friendly.find(nil)        # when you don't know if you even have a slug or ID
+MyModel.friendly.find(nil)        # when you have a variable/param that's possibly nil
 #=> raise ActiveRecord::RecordNotFound
 
-MyModel.friendly.find("bad-slug", allow_nil: true) # where bad-slug is not a valid slug
-MyModel.friendly.find(123, allow_nil: true)        # where 123 is not a valid primary key ID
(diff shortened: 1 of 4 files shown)


lib/friendly_id/finder_methods.rb
```diff
@@ -13,13 +13,16 @@ module FinderMethods
...
+      return super(*args) if args.count != 1 || id.unfriendly_id?
       first_by_friendly_id(id).tap { |result| return result unless result.nil? }
-      return super if potential_primary_key?(id)
+      return super(*args) if potential_primary_key?(id)
 
-      raise_not_found_exception(id)
+      options.symbolize_keys!
+      allow_nil = options.fetch(:allow_nil, false)
+
+      raise_not_found_exception(id) unless allow_nil
```
parndt, line 25:
```suggestion
      raise_not_found_exception(id) unless options.symbolize_keys.fetch(:allow_nil, false)
```

README.md
```diff
@@ -104,6 +104,27 @@ existing users, do this from the console, runner, or add a Rake task:
 User.find_each(&:save)
 ```
 
+## Options
+
+### :allow_nil
```
parndt, line 109:
```suggestion
### `:allow_nil`
```

parndt (approved):

parndt:
Thanks, Joe! 🎉

PR #997, opened 2022-07-15
Fix issues with `allow_nil: true`

joemsak:
Turns out, the documentation I added in #995 wasn't 100% accurate. I had
to add tests for such examples and improve the internal code to support
the option. This should now provide the fully expected behavior.

### DIFF
--- a/lib/friendly_id/finder_methods.rb
+++ b/lib/friendly_id/finder_methods.rb
@@ -27,22 +27,13 @@ module FriendlyId
     # @raise ActiveRecord::RecordNotFound
     def find(*args, allow_nil: false)
       id = args.first
-
-      begin
-        return super(*args) if args.count != 1 || id.unfriendly_id?
-      rescue => e
-        return allow_nil ? nil : raise(e)
-      end
-
+      return super(*args) if args.count != 1 || id.unfriendly_id?
       first_by_friendly_id(id).tap { |result| return result unless result.nil? }
+      return super(*args) if potential_primary_key?(id)
 
-      begin
-        return super(*args) if potential_primary_key?(id)
-      rescue => e
-        return allow_nil ? nil : raise(e)
-      end
-
-      raise_not_found_exception(id) unless allow_nil
+      raise_not_found_exception(id)
+    rescue => e
+      allow_nil ? nil : raise(e)
     end
 
     # Returns true if a record with the given id exists.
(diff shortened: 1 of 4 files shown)


joemsak:
@parndt I missed some necessary changes, this should resolve that.

lib/friendly_id/finder_methods.rb
```diff
@@ -27,9 +27,20 @@ module FinderMethods
...
+      end
+
       first_by_friendly_id(id).tap { |result| return result unless result.nil? }
-      return super(*args) if potential_primary_key?(id)
+
+      begin
+        return super(*args) if potential_primary_key?(id)
+      rescue => e
+        return allow_nil ? nil : raise(e)
+      end
```
joemsak, line 43: I know we talked about exception control flow and I'm definitely not a fan, however I really am not aware of another strategy to use here. I'm open to suggestions though.

lib/friendly_id/finder_methods.rb
```diff
@@ -31,7 +31,9 @@ def find(*args, allow_nil: false)
       first_by_friendly_id(id).tap { |result| return result unless result.nil? }
       return super(*args) if potential_primary_key?(id)
 
-      raise_not_found_exception(id) unless allow_nil
+      raise_not_found_exception(id)
+    rescue => e
+      allow_nil ? nil : raise(e)
```
joemsak, line 36: I know we talked about exception control flow and I'm definitely not a fan, however I really am not aware of another strategy to use here. I'm open to suggestions though.

joemsak:
Dearest @parndt what is the likelihood of this getting merged and then a release? 🙏This PR satisfies the documentation. Let me know if you have feedback. Thanks!

lib/friendly_id/finder_methods.rb
parndt, line 36:
```suggestion
      raise_not_found_exception(id) unless allow_nil
    rescue ActiveRecord::RecordNotFound => exception
      raise exception unless allow_nil
```

What about this version? If we avoid raising the exception in the first place then we get higher performance by avoiding a needless raise/rescue/raise cycle.

parndt:
@joemsak I have been thinking about this change on and off all week! I've added a suggestion, let me know what you think

parndt:
I also spotted a duplicate `raise` which I've removed

joemsak:
Thank you!

parndt (approved):

joemsak:
Hey @parndt any chance on this making it to a release? Let me know your thoughts, thanks!

parndt:
@joemsak yes, definitely, I will get to it as soon as I can. 😄

parndt:
Version 5.5.0 pushed @joemsak

PR #1020, opened 2023-11-11
Fix: history not using `parse_friendly_id`

### LINKED ISSUE #1019: History does not use parse_friendly_id
If `parse_friendly_id` is used, say to downcase, it's used by the regular finder methods, but not by history:

https://github.com/norman/friendly_id/blob/master/lib/friendly_id/history.rb#L94

```ruby
# history.rb:94   
def first_by_friendly_id(id)
  super || slug_table_record(id)
end
```

`super` calls [the finder method](https://github.com/norman/friendly_id/blob/master/lib/friendly_id/finder_methods.rb#L78) which uses `parse_friendly_id`, but if that fails (not found, nil), `slug_table_record` is called with the unparsed ID which results in a query something like this, which will fail because the slug isn't parsed and downcased.

```sql
SELECT ... WHERE  "friendly_id_slugs"."sluggable_type" = 'MyModel' AND "friendly_id_slugs"."slug" = 'UNPARSED-VALUE'
```

Changing the method in `history.rb` to use the parse method results in expected behaviour:

```ruby
# history.rb:94   
def first_by_friendly_id(id)
  super || slug_table_record(parse_friendly_id(id))
end
```

It likely needs to be applied elsewhere, such as `exists_by_friendly_id?`, where the same pattern `super || method_using_unparsed_id(id)` exposes the bug.

I also couldn't get parse_friendly_id to work in the documented way without monkey patching it in the initializer, but that's a different issue.

manafire:
Fixes #1019

### DIFF
--- a/lib/friendly_id/history.rb
+++ b/lib/friendly_id/history.rb
@@ -85,13 +85,13 @@ module FriendlyId
       include ::FriendlyId::FinderMethods
 
       def exists_by_friendly_id?(id)
-        super || joins(:slugs).where(slug_history_clause(id)).exists?
+        super || joins(:slugs).where(slug_history_clause(parse_friendly_id(id))).exists?
       end
 
       private
 
       def first_by_friendly_id(id)
-        super || slug_table_record(id)
+        super || slug_table_record(parse_friendly_id(id))
       end
 
       def slug_table_record(id)


parndt (approved):

parndt:
Thanks @manafire !

parndt:
@manafire I've fixed CI, can you please pull in the latest to your branch so that CI will pass?

manafire:
@parndt Thanks - done!

parndt:
Thank you!

PR #1027, opened 2024-11-20
Fix up transliteration documentation.

bjhess:
The current documentation for transliteration does not transliterate.

```ruby
> "для Минималистов".to_slug.normalize!(:transliterations => [:russian, :latin]).to_s
=> "для-минималистов"
```

These two commands do successfully transliterate, so let's update the documentation to the simplest of the two:

```ruby
> "для Минималистов".to_slug.normalize!(:transliterate => [:russian, :latin]).to_s
=> "dlya-minimalistov"
> "для Минималистов".to_slug.normalize(:transliterate => true, :transliterations => [:russian, :latin]).to_s
=> "dlya-minimalistov"
```

### DIFF
--- a/lib/friendly_id/base.rb
+++ b/lib/friendly_id/base.rb
@@ -150,7 +150,7 @@ module FriendlyId
     #       config.use :slugged
     #       config.use Module.new {
     #         def normalize_friendly_id(text)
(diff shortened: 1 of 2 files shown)


parndt:
Thank you for this, sorry it took so long to merge.

PR #1035, opened 2025-08-04
enhancement: array finders

### LINKED ISSUE #1018: Rails 7.1 composite key support
using rails 7.1 with composite + freindly_id will give the following error

```ruby
extend FriendlyId
```

```
`scope_for_slug_generator': ["column_one", "column_two"] is not a symbol nor a string (TypeError)
scope.where(self.class.base_class.arel_table[primary_key_name].not_eq(send(primary_key_name)))

```
normally the 
```ruby
primary_key_name = self.class.primary_key
```
would return a string but with composite keys it will return an array.

so above example is
```ruby
self.class.primary_key  
 ["column_one", "column_two"]
 ```

### LINKED ISSUE #1034: Array support for finders
I'm using `friendly_id :name, use: [:slugged, :finders]` inside the `GlossaryTerm` model.

Normal ActiveRecord queries work fine, like:

```ruby
GlossaryTerm.find(1)
GlossaryTerm.find([1])
```

Finders using slugs also work fine:

```ruby
GlossaryTerm.find("one")
```

However, when an array of slugs is passed in, the following error is raised:

```ruby
GlossaryTerm.find(["one"])
# => ActiveRecord::RecordNotFound: Couldn't find GlossaryTerm with 'id'=one
# from /home/bob/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/activerecord-8.0.2/lib/active_record/relation/finder_methods.rb:428:in `raise_record_not_found_exception!'
```

Would it be possible to add support for finding records by an array of slugs, similar to how ActiveRecord handles an array of IDs?

This would be helpful.

Paul-Bob:
Fixes #1034 
Possibly fixes #1018  (not confirmed)

## Description

This PR adds support for finding records using an array of slugs, similar to how ActiveRecord handles arrays of numeric IDs. If FriendlyId is the "Swiss Army bulldozer" of slugging plugins, this enhancement adds laser-guided precision to that bulldozer!

## Problem

Previously, while single slug lookups worked fine:
```ruby
GlossaryTerm.find("one")        # ✅ Works
GlossaryTerm.find([1, 2])       # ✅ Works (numeric IDs)
```

Array slug lookups would fail:
```ruby
GlossaryTerm.find(["one"])      # ❌ ActiveRecord::RecordNotFound
GlossaryTerm.find(["one", "two"]) # ❌ ActiveRecord::RecordNotFound
```

## Solution

Now you can find records using:

- **Array of slugs**: `GlossaryTerm.find(["one", "two"])`
- **Mixed arrays**: `GlossaryTerm.find(["one", 2, "three"])` (slugs + numeric IDs)
- **Partial results**: `GlossaryTerm.find(["one", "missing"], allow_nil: true)`
- **Relations**: `GlossaryTerm.active.find(["one", "two"])`

## Key Features

- Maintains backward compatibility
- Proper error handling with `ActiveRecord::RecordNotFound`
- Supports `allow_nil: true` for partial results
- Works with scoped relations
- Handles empty arrays gracefully
- Efficient queries (separate slug and ID lookups)

## Testing

Added comprehensive test coverage including:
- Array slug finding
- Mixed array handling
- Error cases
- Edge cases (empty arrays, single elements)
- Relation scoping
- All existing tests continue to pass

This enhancement makes FriendlyId even more developer-friendly while maintaining it's robustness.

### DIFF
--- a/lib/friendly_id/finder_methods.rb
+++ b/lib/friendly_id/finder_methods.rb
@@ -1,3 +1,5 @@
+require "set"
+
 module FriendlyId
   module FinderMethods
     # Finds a record using the given id.
@@ -27,6 +29,12 @@ module FriendlyId
     # @raise ActiveRecord::RecordNotFound
     def find(*args, allow_nil: false)
       id = args.first
+      
+      # Handle array of IDs (potentially including slugs)
+      if args.count == 1 && id.is_a?(Array)
+        return find_by_array(id, allow_nil: allow_nil)
+      end
+      
       return super(*args) if args.count != 1 || id.unfriendly_id?
       first_by_friendly_id(id).tap { |result| return result unless result.nil? }
       return super(*args) if potential_primary_key?(id)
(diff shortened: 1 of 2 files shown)


Paul-Bob:
Hi @norman, thank you for the amazing work on this gem!

I would appreciate a review on this PR. Could you please let me know if there is a chance for this to get merged, or if I should switch to a forked solution instead?

adrianthedev:
Hey @norman.
Thanks for chiming in. I appreciate the context.

I am Adrian, the author of [Avo](https://avohq.io) and a teammate of @Paul-Bob.
I suggested to him to PR this change.

Let me give you some context into why we would love to have this change and why we probably can't do what you suggest very easily.

Avo is a bring-your-own library kind of situation. We don't have `friendly_id` baked in as we don't want the user to have to have a library that they are not using.
They choose to use it (and we use it extensively), so it's difficult for us to write a plugin and confidently require it.
It's wasteful to require a plugin gem for a library that they might not use.

Regarding the proposed change (, it's again something out of our control.
We don't directly instrument friendly_id under the hood. The user does.
And we are forced to give out instruction in our documentation (sample below) which we both know nobody reads and then it kind of falls into one of "Avo is broken" or "Avo doesn't work nicely with friendly_id" buckets.

```ruby
# we tell them to add this bit of code in resources which use friendly_id
# we'd rather have it "just work"
self.find_record_method = -> {
  if id.is_a?(Array)
    Post.where(slug: id)
  else
    Post.find(id)
  end
}
```

So yeah, this would improve compatibility with existing libraries which use the `find` query no matter which underlying slugging library is used, it preserves backward compatibility while enabling flexibility, and it reduces misattributed frustration from downstream developers which is more important in my view.

Personally, I'd love to be able to do things even if they are less performant than to have to jump through hoops to implement something to patch the original behavior.

Thanks for listening, and we are open to other suggestions as well.

parndt:
I've gone back and forth on this but I'm going to close this one. I had a proper re-read through the code and, with Norman's advice about what we used to do, have found a few things that block this:

1. It doesn't work with `:history`. `find_by_array` queries the slug column directly, so it never calls `first_by_friendly_id`, which is the method `History::FinderMethods` overrides so, on a model using history, `find(old_slug)` works but `find([old_slug])` raises.
2. String IDs raise even when the records are there. Anything from params is a string, so `numeric_ids` holds `"1"` while `found_ids` holds `1`. The subtraction never cancels out and the missing-ID check fires. `find([1])` works, `find(["1"])` doesn't.
3. Slugs that look like numbers stop working. Normally find tries the slug first and only falls back to the primary key if that misses. The array path does it the other way round, so a record slugged `"2024"` is fine with `find("2024")` but not `find(["2024"])`.

The other thing is that Avo has since solved this itself nicely in avo-hq/avo#4012 which added friendly_id detection by default, with avo-hq/avo#4122 following up for scoped models. So Avo users get friendly finds without setting anything up, which is what you were after, I think? I believe that's the better place for it: Avo knows it's dealing with a list of selected rows, and find here stays about finding one record.

For what it's worth, if you ever want this in your own code, this works today:

```ruby
ids.map { Model.friendly.find(it) }
```

It's a query per ID, but it handles history, string IDs, numeric-looking slugs, and any custom `parse_friendly_id`, because each one goes through the normal path.

Thanks so much, everyone.

PR #1037, opened 2025-09-26
Add treat_numeric_as_conflict option to prevent ambiguous numeric slugs

oehlschl:
## Summary

We sometimes have issues where clients call our APIs and pass integer primary keys instead of slugs, but `Model.friendly.find(id)` finds a record with the same numeric slug, because our slug candidates are dynamic and in some cases purely numeric.  

We'd like to prevent that, and so this PR adds a new configuration option `treat_numeric_as_conflict` prevent purely numeric slugs and therefore resolve the ambiguity where `Model.friendly.find("123")` could return either a record with `slug="123"` OR a record with `id=123`.

When enabled, purely numeric slug candidates are treated as conflicts and resolved using UUID suffixing, ensuring all slugs contain non-numeric characters that eliminate lookup ambiguity.

## Changes

- Added `treat_numeric_as_conflict` configuration option to `Slugged::Configuration`  
- Updated `SlugGenerator` to detect and reject purely numeric slugs when enabled
- Added `purely_numeric_slug?` helper method using `Integer()` validation (consistent with `potential_primary_key?`)
- Expanded test coverage for both default behavior and numeric prevention

## Usage

```ruby
class Product < ApplicationRecord
  extend FriendlyId
  friendly_id_config.treat_numeric_as_conflict = true
end

# Results:
Product.create(name: "123")     # slug: "123-f9f3789a-daec-4156-af1d-fab81aa16ee5"
Product.create(name: "abc123")  # slug: "abc123" (unchanged)
```

## Testing Done

- All existing tests pass
   - Verified backward compatibility (default behavior unchanged)
- Added comprehensive test coverage for new functionality
   - Tests cover edge cases: zero, large numbers, alphanumeric combinations
   - Verified UUID conflict resolution works for multiple numeric slugs

## Deploy Notes

- **No breaking changes**: This is a new opt-in configuration option
- **No migration required**: Uses existing slug storage and UUID conflict resolution
- **Safe to deploy**: Feature is disabled by default, maintaining all existing behavior
- **Rollback safe**: Simply remove configuration option to disable

### DIFF
--- a/lib/friendly_id/slug_generator.rb
+++ b/lib/friendly_id/slug_generator.rb
@@ -12,12 +12,28 @@ module FriendlyId
         return false if @config.reserved_words.include?(slug)
       end
 
+      if @config.treat_numeric_as_conflict && purely_numeric_slug?(slug)
+        return false
+      end
+
       !@scope.exists_by_friendly_id?(slug)
     end
 
     def generate(candidates)
       candidates.each { |c| return c if available?(c) }
       nil
+    end
+
+    private
+
+    def purely_numeric_slug?(slug)
+      return false unless slug
+      begin
+        Integer(slug, 10)
+        slug.to_s == Integer(slug, 10).to_s
+      rescue
+        false
+      end
     end
   end
 end
--- a/lib/friendly_id/slugged.rb
+++ b/lib/friendly_id/slugged.rb
@@ -386,7 +386,7 @@ module FriendlyId
     # {FriendlyId::Configuration FriendlyId::Configuration}.
     module Configuration
       attr_writer :slug_column, :slug_limit, :sequence_separator
-      attr_accessor :slug_generator_class
+      attr_accessor :slug_generator_class, :treat_numeric_as_conflict
 
       # Makes FriendlyId use the slug column for querying.
       # @return String The slug column.
--- a/test/numeric_slug_test.rb
+++ b/test/numeric_slug_test.rb
@@ -1,5 +1,10 @@
 require "helper"
 
+class Article < ActiveRecord::Base
+  extend FriendlyId
+  friendly_id :name, use: :slugged
+end
+
 class NumericSlugTest < TestCaseClass
   include FriendlyId::Test
   include FriendlyId::Test::Shared::Core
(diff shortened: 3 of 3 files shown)


oehlschl:
@parndt I know there's not a ton of active development on friendly_id now, but any concerns with something like this? Thanks in advance.

oehlschl:
@parndt I added Guide documentation, and this change to the changelog (under Unreleased). Also updated the Int casting (will comment inline); please let me know if you need anything else. Thanks!

lib/friendly_id/slug_generator.rb
```diff
@@ -12,12 +12,28 @@ def available?(slug)
...
     end
+
+    private
+
+    def purely_numeric_slug?(slug)
+      return false unless slug
+      begin
+        Integer(slug, 10)
+        slug.to_s == Integer(slug, 10).to_s
+      rescue
```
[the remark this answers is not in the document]
oehlschl: `Integer` can raise `ArgumentError` or `TypeError`, so I rescued those; note that there are existing call sites with open rescues, with I don't think is uncommon for type casting, but I'm happy to do either. https://github.com/norman/friendly_id/blob/c288abb863cc0ad3a57131f1047542969776ecb7/lib/friendly_id/finder_methods.rb#L67

I did fix the duplicate casting, though, so thanks for that!

lib/friendly_id/slug_generator.rb
```diff
@@ -12,12 +12,27 @@ def available?(slug)
         return false if @config.reserved_words.include?(slug)
       end
 
+      if @config.treat_numeric_as_conflict && purely_numeric_slug?(slug)
```
parndt: @oehlschl I just had one last question: if this is only available when using `slugged` (as your docs indicate) then should we check here for `if @config.uses?(::FriendlyId::Slugged)` as well?

oehlschl: In other words, it didn't need seem necessary to add since this whole class depends on `Slugged`. But let me know if that's not the case.

parndt: oh, of course! Thank you

parndt:
@oehlschl I will work on it; I've been meaning to set up trusted publishing for a while..

parndt:
@oehlschl after much ado with rubygems publishing, we have [version 5.6.0](https://rubygems.org/gems/friendly_id/versions/5.6.0)!

PR #1046, opened 2026-07-13
Treat leading-zero slugs as numeric conflicts

fschwahn:
Previously, numeric detection parsed the slug as an integer and compared the result, causing values such as `00123` to be treated as non-numeric. The check now matches ASCII digits directly.

Includes a regression test for leading-zero numeric slugs. (I ran into this issue when an international phone number was accidentally use as a slug)

### DIFF
--- a/lib/friendly_id/slug_generator.rb
+++ b/lib/friendly_id/slug_generator.rb
@@ -28,11 +28,8 @@ module FriendlyId
 
     def purely_numeric_slug?(slug)
       return false unless slug
-      begin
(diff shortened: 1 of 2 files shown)


fschwahn:
@parndt slight bump - the current behavior is still producing numeric slugs for us in these edge cases - it would be great if this could be merged. I believe the CI failures are unrelated to these changes.

parndt:
@fschwahn thanks, CI should be resolved if you rebase latest

fschwahn:
I see you already added the changelog entry. Thank you very much!

